PR #42, opened 2025-03-07
fix: better prompt for resolution tracking

binary-koan:


apps/nextjs/src/inngest/functions/checkConversationResolution.ts
```diff
@@ -8,15 +8,15 @@ import { runAIQuery } from "@/lib/ai";
 import { loadPreviousMessages } from "@/lib/ai/chat";
 import { GPT_4O_MINI_MODEL } from "@/lib/ai/core";
 
-const RESOLUTION_CHECK_PROMPT = `You are analyzing a customer service conversation to determine if the customer's issue was resolved.
+const RESOLUTION_CHECK_PROMPT = `You are analyzing a customer service conversation to determine if the customer's issue was addressed.
```
binary-koan, line 11: It seems to become quite strict if you talk about "resolved" and doesn't count resolutions where the AI explained but didn't specifically do something to solve the problem

PR #54, opened 2025-03-12
feat: inline command bar

binary-koan:
Part of #3 

Changes the command bar to open inline instead of as a full screen modal, and generally moves closer to the new design.

Also restored the missing reply buttons on mobile.

https://github.com/user-attachments/assets/fc346fee-b7b5-4241-a647-b74337a96595


binary-koan:
@slavingia Is this what you imagined in terms of UX?

slavingia:
Yep! Would be nice to have suggestions underneath it so they're exposed right away, but we can ship that next.

binary-koan:
Ah yeah meant to say that suggestions will be the next PR

Issue #68, opened 2025-03-14
Support "chatting" via email


binary-koan:
- [ ] Remove the "continue your conversation ..." email
- [ ] When receiving an email, use the same chat AI to generate and email back a response
    - [ ] Consolidate the logic in one place
- [ ] Handle tools correctly (e.g. screenshots are only available in chat)

### Why

Easier for people to get help (removes the barrier of logging in and looking for the notification on the website). There's no good reason to have two different paths for email and chat, having one path for all messages will streamline the logic.

binary-koan:
Should we add something in the email to clarify it's an AI response, or even some kind of "let us know if you need human assistance"? With email it's not as obvious that an AI is answering which might turn people off (although ideally the AI will do a good enough job that they won't mind!)

![Image](https://github.com/user-attachments/assets/0367cc60-6e3f-4c5f-af6a-ac842d5c0b55)

(@raphaelcosta @Rachel-splatt lmk if you have thoughts)

raphaelcosta:
> Should we add something in the email to clarify it's an AI response, or even some kind of "let us know if you need human assistance"? With email it's not as obvious that an AI is answering which might turn people off (although ideally the AI will do a good enough job that they won't mind!)

I believe it’s a good practice to always indicate that this is an AI-generated response. We can also guide users on what they should say to obtain human assistance.

Rachel-splatt:
I agree! Could be something like:

"This response was created by our AI support agent. Need human support? Let us know in your reply." 

Or if needed we could instruct them to reply with a specific word or phrase: "...Need human support? Reply with 'HUMAN' to connect with a team member."

binary-koan:
> "This response was created by our AI support agent. Need human support? Let us know in your reply."

This is perfect, will add it!

PR #71, opened 2025-03-15
feat: floating editor toolbar

binary-koan:
Part of #3, [ref](https://anti--work.slack.com/archives/C055QQCQFM1/p1740584891555409?thread_ts=1740509649.981519&cid=C055QQCQFM1)

Updates the editor to use a floating bottom-aligned toolbar which is collapsible to save space (similar idea to Gmail)

Screenshot 2025-03-15 at 23 07 47

Screenshot 2025-03-15 at 23 09 01

The only thing missing from the design is the signature, which isn't relevant yet since we don't have signatures in that way right now. Will add that as part of #68


binary-koan:
@raphaelcosta _a wdyt about this? I don't think we can hide the toolbar when text isn't selected because adding attachments and using lists etc isn't intuitive that way (if we really want that then IMO we should go full Medium/Notion and have a "+" button on blank lines, but a manually collapsible toolbar seems much easier)

raphaelcosta (approved):
@binary-koan _a looks good!

Issue #81, opened 2025-03-18
Redirection from dashboard to mailbox fails to load the conversation


khalatevarun:
## Issue
* On redirecting from the dashboard to mailbox the conversation fails to load
<img width="885" alt="Image" src="https://github.com/user-attachments/assets/66992355-3c46-4a0e-b2ff-e16b7f61ec98" />

## Cause
* The link associated with the conversation in the dashboard are - /conversations?id={message.id}-message
* The mail box page fails to load this. Instead the slug should be - /conversation?id={message.slug}

PR #82, opened 2025-03-18
fix: update conversation links in RealtimeEvents and GitHub router

vikaswakde:
fix : #81 

### Description: 

- Changed conversation ID to conversationSlug in RealtimeEvents component link.
- Updated body text in GitHub router to use conversationSlug instead of conversation ID for consistency.

### Screenshot 

https://github.com/user-attachments/assets/7c82a2e9-6032-4471-a430-4897b49f7015


binary-koan (approved):
Nice catch, thanks!

PR #95, opened 2025-03-20
fix: remove PageContainer component to simplify layouts

binary-koan:
It was written as a workaround because we assumed pages would follow the horizontal inbox layout by default, but that's not the case.


raphaelcosta:
It looks like this PR removed the vertical scroll from the Settings page

binary-koan:
True, checking

PR #96, opened 2025-03-20
feat: buttons to ask for human support and give feedback

### LINKED ISSUE #90: All tickets should lead to a human
1. AI response 
2. Button to escalate to a human upon every response: "I need additional help. I'd like to talk to a human for further help."

Something like that! Please prioritize asap so all creators can get human support (and know they can)

raphaelcosta:
- Add a button for requesting human support, so users don’t have to guess how to do it.  
- Since we’re now adding action buttons, also include one to make it easier for users to track resolution updates.  
- Resolves #90 

https://github.com/user-attachments/assets/bf388555-d718-4573-844b-48432f3c6c9c


raphaelcosta (commented):
@binary-koan _a could you review this one?

binary-koan (commented):
@raphaelcosta _a Isn't the idea to escalate immediately if the button is clicked? (I like the idea of asking for more details if it's unclear but the "I'll either assist or escalate" response makes the button feel misleading - maybe solvable just with better prompting)

raphaelcosta:
> @raphaelcosta _a Isn't the idea to escalate immediately if the button is clicked? (I like the idea of asking for more details if it's unclear but the "I'll either assist or escalate" response makes the button feel misleading - maybe solvable just with better prompting)

I’m thinking of adding something to the tool prompt so it automatically triggers the tool when that exact phrase was received. I didn’t do it here to avoid conflicts with the other PR, but I believe it simplifies the logic quite a bit. If it doesn’t work well, we can always change it later to call the escalation function.

raphaelcosta:
@binary-koan _a I’ve adjusted the prompt so the conversation gets escalated right away, even if there isn’t a specific question yet

slavingia:
It's a bit weird imo to have four possible buttons, two that are effectively it solved / thumbs up, and two that are bad / talk to a human. Should we merge?

raphaelcosta:
> It's a bit weird imo to have four possible buttons, two that are effectively it solved / thumbs up, and two that are bad / talk to a human. Should we merge?

Yeah, that makes sense. Right now, "it’s solved" and thumbs up are basically the same action. We could also treat "talk to a human" as a negative reaction and remove the thumbs up/down buttons below each message. WDYT?

slavingia:
Maybe worth a couple others thoughts but I think so; merging them makes sense to me.

binary-koan:
I agree with merging the buttons. Not sure that we should treat "talk to a human" as a negative reaction since there are cases that genuinely should be escalated; IMO it's fine if we just have users marking as positive + staff marking as negative based on escalations

PR #101, opened 2025-03-22
feat: true native Apple auth in Tauri

binary-koan:
Apple seems to have changed (broken) the Safari behaviour we relied on for native auth before, and are rejecting our builds because the Apple sign in is not seamless.

This adds actual native sign-in using Rust Obj-C bindings.

Screenshot 2025-03-22 at 23 58 10


apps/nextjs/src/trpc/router/user.ts
```diff
@@ -10,4 +13,46 @@ export const userRouter = {
...
+
+      const idToken = await appleSignin.verifyIdToken(tokenResponse.id_token, {
+        audience: env.APPLE_APP_ID,
+      });
+
+      const users = await clerkClient.users.getUserList({ emailAddress: [idToken.email] });
+
+      let user = users.data[0];
+      if (!user) {
+        user = await clerkClient.users.createUser({
```
binary-koan, line 45: Clerk doesn't seem to have a way to authenticate based on an existing authorization code. Will check with them but this works for now

apps/tauri/src-tauri/src/lib.rs
```diff
@@ -30,15 +29,58 @@ fn start_options() -> serde_json::Value {
...
+}
+
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
     let mut builder = tauri::Builder::default()
-        .plugin(tauri_plugin_log::Builder::new().build())
+        .plugin(
+            tauri_plugin_log::Builder::new()
+                .target(tauri_plugin_log::Target::new(
+                    tauri_plugin_log::TargetKind::Webview,
```
binary-koan, line 70: Allows us to call `__TAURI__.log.attachConsole()` in dev tools to get Tauri logs in the JS console - useful for Mac App Store where we don't necessarily have permission to write to the standard log file location

Issue #102, opened 2025-03-23
Search Page: List Item Content Overflows Container


khalatevarun:
On the search results page, the content inside each list item overflows its container instead of wrapping or truncating properly

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/056237ef-2a59-4ee0-9db4-ec3badda9bbe" />

PR #103, opened 2025-03-23
fix: list item content overflow in search results

khalatevarun:
fix: #102 

### Before
image

### After
image


binary-koan (approved):
I just ran into this too, thanks for the fix!

Issue #121, opened 2025-03-26
Intermittent timeouts when loading pages


binary-koan:
[ref](https://anti--work.slack.com/archives/C055QQCQFM1/p1742172305398259)

We're occasionally getting timeouts when loading the conversation list in production, with crazy long tRPC times (24 seconds for `mailbox.list` in one example I found). Happens anywhere from once every few days to multiple times a day.

raphaelcosta:
Adding priority to this as it may affect our demos

raphaelcosta:
I am fixing the hard delete function to remove conversations from inactive organizations and check whether it reduces DB CPU usage

raphaelcosta:
We will probably need to reduce the dimensions of the embeddings. I will check it tomorrow.

raphaelcosta:
It looks like database consumption was the main issue, and it's now improved. We've also enhanced our performance monitoring and integrated Sentry for APM.  Should we close this issue @binary-koan ?
<img width="1149" alt="Image" src="https://github.com/user-attachments/assets/278d97a6-2b07-4ba6-8705-8025c43ebf07" />

binary-koan:
Yep let's close and investigate further if it reoccurs

PR #122, opened 2025-03-26
feat: benchmark Clerk API calls

binary-koan:
Related to #121 

One of my suspicions is that Clerk API calls may be slow; this adds logging to give an initial idea of whether they're an issue in the request timeout logs.


binary-koan:
Yeah worth considering. Aside from this I've seen the frontend redirecting to and from Clerk URLs before loading the actual page (re-authenticating for security or something?) which makes everything feel slower + I agree with making setup easier for contributors

Not sure if we've reached the tipping point yet but the reasons are piling up

PR #123, opened 2025-03-26
chore: make autofix work without Vercel so it runs on contributor PRs

binary-koan:
Updates `.env.local.sample` to have valid values for validated env vars so that we can use it in autofix instead of pulling from Vercel.

At the moment it takes time for one of us to run autofix manually on contributor PRs.


apps/nextjs/.env.local.sample
```diff
@@ -1,5 +1,4 @@
 ABLY_API_KEY="set_me"
-ADDITIONAL_PAID_ORGANIZATION_IDS="set_me"
```
binary-koan, line 2: Will add this back commented as part of making more env vars optional

Issue #129, opened 2025-03-27
db:reset script fails for fresh setup


naz:
The expected behavior for db:reset is to work out of the box.

Currently, running `npm run db:reset` fails with:
```
Database reset failed: Error: CLERK_INITIAL_ORGANIZATION_ID and CLERK_INITIAL_USER_IDS must be set for seeds to run.
```

The right approach would be to have both `CLERK_INITIAL_ORGANIZATION_ID` and `CLERK_INITIAL_USER_IDS` present in the `.env.local.sample` with `"set_me"` values.

The [env.ts has the validations](https://github.com/antiwork/helper/blob/7af7b781eb03f792ded3f82109589ef1b4c42ff6/apps/nextjs/src/env.ts#L50-L57) for these keys already, so just need to add them in the sample env file.

### Why

When a new developer sets up the environment at minimum the `.env.local` file should contain all required variables one needs to set.

raphaelcosta:
Closing this since it was intentional to not include the ENV vars. Adding them to the sample file could be misleading, but we’d be happy to accept a PR that clarifies this in the documentation or in the error message. Comment for reference: https://github.com/antiwork/helper/pull/130#pullrequestreview-2720894282

PR #130, opened 2025-03-27
fix: added missing env variables

naz:
refs https://github.com/antiwork/helper/issues/129

Adding these keys improves the developer experience when setting up the project


binary-koan (commented):
@naz _a Not including them was deliberate since you can't log in after `db:reset` anyway unless you sign up for Clerk and add valid values from your instance. Maybe clarifying that in the error message and/or checking that the org and users can be fetched with the Clerk API would be a better solution - what do you think?

(there's an ongoing conversation about whether we should replace Clerk for precisely this reason btw)

PR #139, opened 2025-03-27
feat: use pg instead of postgres

binary-koan:
This should allow us to capture query spans in Sentry with their built-in instrumentation.


binary-koan (commented):
@raphaelcosta _a Could you double check? Seems to work on preview, hopefully won't have problems on production

I didn't manage to test Sentry locally (it didn't show any "development" transactions even when I turned everything on) but this will hopefully add query spans out of the box on production

raphaelcosta (approved):
@binary-koan Looks good! Thanks

PR #144, opened 2025-03-28
Marketing page card content update

Rachel-splatt:
## What 

Update the content of marketing page cards, since style linter and FAQs no longer exist.

Screenshot 2025-03-28 at 12 16 42 AM

Open to feedback on what we should highlight here instead.


slavingia:
If we don't highlight tool usage that would be nice in a card.

Rachel-splatt:
Sounds good! Replaced the in-app chat card with tools, since in-app chat is already highlighted with the widget.

Screenshot 2025-03-30 at 10 11 31 PM

PR #148, opened 2025-03-28
fix: make email auto-response trigger more reliably

binary-koan:
[ref](https://anti--work.slack.com/archives/C085P30V4LT/p1743166612478799?thread_ts=1743166113.704479&cid=C085P30V4LT)

Moved a bunch of logic into the respond function (which retries) rather than having it in the webhook function (which doesn't). Also updated the returned values to make runs easier to filter on the Inngest dashboard.


binary-koan (commented):
@raphaelcosta _a Could you review? Test failures look unrelated

raphaelcosta (approved):
@binary-koan _a Looks good!

PR #151, opened 2025-03-28
perf: add index for conversations count

raphaelcosta:
- Part of #121 
- Results are from production, and the index has already been added there for testing

| EXPLAIN ANALYZE |
|--------|
| BeforeScreenshot 2025-03-28 at 3 25 51 PM |
| After Screenshot 2025-03-28 at 3 26 03 PM|


raphaelcosta (commented):
@binary-koan _a could you please review?

binary-koan (approved):
@raphaelcosta _a Looks good! Should we remove the status-only index later?

raphaelcosta:
> @raphaelcosta _a Looks good! Should we remove the status-only index later?

Yes, I will do that!

PR #152, opened 2025-03-28
perf: remove left join and reuse existing results from keyword search

raphaelcosta:
Aiming to optimize the performance of [this query](https://gumroad-to.sentry.io/insights/backend/database/spans/span/55ce4a07625d2fe0?environment=vercel-production&project=4508049493196800&spansSort=-avg%28span.self_time%29&statsPeriod=1h), which is currently taking up to 2 seconds.

- Remove the unnecessary left join by leveraging existing results from the keyword search
- Part of #121


raphaelcosta (commented):
@binary-koan _a another for your review

binary-koan (approved):
@raphaelcosta _a LGTM!

PR #156, opened 2025-03-28
fix: exclude merged conversations from counts

binary-koan:
Realised the open count didn't update after I merged existing open conversations.


binary-koan:
@raphaelcosta Just now you added the new index I realised we need to update this query 🤦 Can fix the index again if we need (actually, for another optimisation I don't think we need to return anything other than "open" here - we don't show closed count any more)

raphaelcosta:
> @raphaelcosta Just now you added the new index I realised we need to update this query 🤦 Can fix the index again if we need (actually, for another optimisation I don't think we need to return anything other than "open" here - we don't show closed count any more)

Running that query only for open tickets would be nice!

PR #163, opened 2025-03-30
feat: tabs in desktop app

### LINKED ISSUE #161: Add tabs to desktop app
Something like this:

Image

~Due to a [Clerk limitation](https://feedback.clerk.com/roadmap?id=9a9fe6d6-f4ce-4196-8998-96ed41fa3b81) it's not feasible to have top level Figma-style tabs, but we can at least support opening multiple conversations in the same inbox in tabs.~

Add another WebView for the tab bar in the native window, and set up communication to open and switch tabs.

### Why

So you can view multiple tickets at once in the desktop app and don't need to keep going backwards and forwards (or opening browser tabs) when looking at similar ones etc.

binary-koan:
Closes #161 

Adds a tab bar to the desktop app
* Automatically opens `target="_blank"` links in new tabs
* Tracks recently closed tabs and allows reopening them

Demo (I've since fixed the gap at the top of the settings page):

https://github.com/user-attachments/assets/6640638c-46a1-4e5a-9b4e-d0a012c935c3

Will open an issue in Tauri about fixing the traffic light button position (centering them in the tab bar) - apparently it just became possible but I can't get it to work.


binary-koan:
@slavingia How does this look? (demo in description)

PR #170, opened 2025-03-31
feat: use previous tools results

raphaelcosta:
Add previous tool results to the chat history, allowing one tool to use those results as arguments for subsequent tool calls. For example, use the purchase ID from a previous search result to trigger an auto-refund.


apps/nextjs/src/lib/ai/chat.ts
```diff
@@ -77,7 +77,27 @@ export const loadPreviousMessages = async (conversationId: number, latestMessage
   return conversationMessages
     .filter((message) => message.body && message.id !== latestMessageId)
     .map((message) => {
-      const messageRecord = message as any; // Type assertion to handle union type
+      const messageRecord = message as any;
```
binary-koan, line 80: (While you're in the area, why are we casting this to any?)

PR #177, opened 2025-04-02
feat: simplify required env vars

binary-koan:
* Updates a bunch of environment variables to be optional and disables the associated features if they're not set.
* Reorganises variables in `env.ts` and `env.local.sample` to make it a bit clearer how to set up the project.
* Removes a few vars that are no longer used.


binary-koan (commented):
@raphaelcosta _a Could you review? LMK if you feel any of these optional vars should be required or vice versa

raphaelcosta (approved):
@binary-koan _a looks good!

PR #200, opened 2025-04-07
fix: Masked Sentry DSN and enabled 100% tracing

### LINKED ISSUE #195: Avoid exposing Sentry DSN
apps/nextjs/sentry.common.config.ts

Ideally, Sentry DSN should be an env variable..

Shubhdeep12:
fixed #195


apps/nextjs/.env.local.sample
```diff
@@ -15,6 +15,9 @@ AWS_DEFAULT_REGION="set_me"
 AWS_PRIVATE_STORAGE_BUCKET_NAME="set_me"
 CLERK_SECRET_KEY="set_me"
 NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY="pk_test_yourkeyKJA"
+NEXT_PUBLIC_SENTRY_DSN="set_me"
+SENTRY_DSN="set_me"
```
binary-koan, line 19: Can we just have one env var? Sentry's default setup is to combine frontend and backend errors so I'd rather follow that unless there's a strong reason to split it

Also it can be commented out under "optional integrations" since it's optional

apps/nextjs/sentry.common.config.ts
```diff
@@ -1,11 +1,11 @@
 import { env } from "@/env";
 
 export const commonConfig = {
-  dsn: "https://84640ca2ac12acbfd7a06e0a02e3ba56@o4508048702767104.ingest.us.sentry.io/4508049493196800",
+  dsn: process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN,
   enabled: env.NODE_ENV === "production",
 
   // Define how likely traces are sampled. Adjust this value in production, or use tracesSampler for greater control.
-  tracesSampleRate: 0.1,
+  tracesSampleRate: 1,
```
binary-koan, line 8: This should either be configurable or based on the environment, it's prohibitively expensive to enable it for all requests in production

binary-koan (commented):
@Shubhdeep12 _a Thanks for improving this! Added a couple of comments

apps/nextjs/.env.local.sample
Shubhdeep12: Next.js needs SENTRY_DSN for server-side and NEXT_PUBLIC_SENTRY_DSN for client-side - same value, just different prefixes so Next can expose it properly.
That's why I added both in the common config.

Also yes, I can comment it out since it's optional
thanks!

apps/nextjs/sentry.common.config.ts
Shubhdeep12: Let's revert this for now, then.
We can increase it later if the need arises.

apps/nextjs/.env.local.sample
binary-koan: Can we not just use the NEXT_PUBLIC_ var on the server as well?

Shubhdeep12: I guess yes. we can do that
https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#:~:text=Non%2D,different%20environment).

Shubhdeep12: NEXT_PUBLIC_ is available on both server and client side

apps/nextjs/sentry.common.config.ts
```diff
@@ -1,7 +1,7 @@
 import { env } from "@/env";
 
 export const commonConfig = {
-  dsn: "https://84640ca2ac12acbfd7a06e0a02e3ba56@o4508048702767104.ingest.us.sentry.io/4508049493196800",
+  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
```
binary-koan, line 4:
```suggestion
  dsn: env.NEXT_PUBLIC_SENTRY_DSN,
```

(this will probably get flagged by the linter)

binary-koan (approved):
Looks good, thanks!

PR #218, opened 2025-04-09
fix: improve open count query

raphaelcosta:
- Only return the open count per status, since that’s the only one we display.
- Simplify the logic to reduce both query complexity and the total number of queries.
- Renamed tRPC query from `countByStatus` to `openCount`

### DIFF
--- a/apps/nextjs/src/app/(dashboard)/mailboxes/[mailbox_slug]/(inbox)/_components/list.tsx
+++ b/apps/nextjs/src/app/(dashboard)/mailboxes/[mailbox_slug]/(inbox)/_components/list.tsx
@@ -171,15 +171,12 @@ const ListContent = ({ variant }: { variant: "desktop" | "mobile" }) => {
   }, [fetchNextPage, hasNextPage, isFetchingNextPage]);
 
   const statusOptions = useMemo(() => {
-    const statuses = status.flatMap((status) =>
-      status.status
+    const statuses = status.flatMap((s) =>
+      s.status
         ? {
-            value: status.status,
-            label:
-              status.status === "open"
-                ? `${status.count.toLocaleString()} ${capitalize(status.status)}`
-                : capitalize(status.status),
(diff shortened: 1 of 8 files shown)


apps/nextjs/src/trpc/router/mailbox/index.ts
```diff
@@ -46,32 +46,33 @@ export const mailboxRouter = {
...
+      countOpenStatus(eq(conversations.assignedToClerkId, ctx.session.userId)),
+      countOpenStatus(isNotNull(conversations.assignedToClerkId)),
     ]);
 
     return {
       conversations: all,
       mine,
       assigned,
-      unassigned,
+      unassigned: all - assigned,
```
raphaelcosta, line 75: Another query isn’t necessary, this logic handles it.

raphaelcosta (commented):
@binary-koan _a could you please review?

binary-koan:
@raphaelcosta Will review in an hour or two, but FYI I’m working on optimistically updating that query (in case you were going to do that as well)

raphaelcosta:
> @raphaelcosta Will review in an hour or two, but FYI I’m working on optimistically updating that query (in case you were going to do that as well)

Thanks! All good, I am not working on that, just trying to reduce the complexity when it's actually querying

apps/nextjs/src/app/(dashboard)/mailboxes/[mailbox_slug]/(inbox)/_components/list.tsx
```diff
@@ -171,27 +171,22 @@ const ListContent = ({ variant }: { variant: "desktop" | "mobile" }) => {
   }, [fetchNextPage, hasNextPage, isFetchingNextPage]);
 
   const statusOptions = useMemo(() => {
-    const statuses = status.flatMap((status) =>
-      status.status
+    const statuses = status.flatMap((s) =>
+      s.status
```
Comment (binary-koan) [inline apps/nextjs/src/app/(dashboard)/mailboxes/[mailbox_slug]/(inbox)/_components/list.tsx:175]: Isn't this always defined now?

PR #219, opened 2025-04-09
fix: request cache clerk calls

raphaelcosta:
- Cache other Clerk calls in the same request
- Remove unncessary Clerk call that is fetching the organization


raphaelcosta (commented):
@binary-koan _a could you please review and check if you agree with those changes?

binary-koan (approved):
@raphaelcosta _a Makes sense to me!

Issue #51, opened 2025-03-10
Helper agent


slavingia:
Command Helper from within Slack.

e.g. @helper give me 5 tickets to respond to, about PayPal

Actions:
- Assign me 5 tickets
- Basic queries (how many tickets have I answered in the last 24 hours?)
- Close all open tickets older than 30 days about PayPal
- Reply to all open tickets regarding identity verification saying something along the lines of: "This issue has been resolved!" - but customize for each

- [ ] Make search smarter to support the above

We should use the Vercel AI SDK. Also, refer to this Vercel AI SDK guide on building a Slack bot: https://sdk.vercel.ai/docs/guides/slackbot

slavingia:
Think this will be hugely important. Slack will become the default interface for most SaaS, I expect.

skushagra9:
Hey @slavingia , 
Super excited to pick this up.

slavingia:
No need to signal your intent before opening the PR for feedback, unless you have specific questions.

vikaswakde:
@slavingia suppose we ask AI agent to close tickets about "PayPal" should it search across all the mailboxes ? and close all of them?

slavingia:
Good point. It should ask for the mailbox if not specific and take that as a required param for tool usage if there's more than one mailbox for the org imo.

vikaswakde:
@slavingia what about this : Close all open tickets older then 30 days about PayPal (should) ai target and close all tickets with that are related to email with this keyword PayPal" ? or  should it search for word PayPal in ticket subject ? because, if this is the case, then all the tickets with "paypal" will  be closed by our AI, 

what exact condition should AI close on to reduce False Positives ?

because what I currently was doing was, searching though subjects/title that contain PayPal and close those issues, but it closing all the tickets with that keyword, 

What do you think ?

slavingia:
It should use AI to make the determination - "does this meet the criteria" - in the case of "about PayPal" most/all tickets that contain that word should

Unless there's a ticket "this isn't about PayPal" then it shouldn't!

vikaswakde:
@slavingia checkout https://github.com/antiwork/helper/pull/74

binary-koan:
@raphaelcosta Is it worth adding MCP and running the agent through that? (not 100% up to speed on MCP and agents in general yet, but seems like we'd effectively want that for the Slack agent anyway)

binary-koan:
[...]

PR #224, opened 2025-04-09
feat: Slack agent

binary-koan:
Part of #51 

Adds a Slack agent which can fetch ticket and member information. This is a relatively basic implementation so far but it sets us up to add more tools.

Tools not included from the issue description
* Close all tickets matching a search (need to consider pagination, limits, etc)
* Reply to all tickets matching a search (should have some kind of confirmation IMO and need to find a way to search tickets "about identity verification")

Also added a debug mode (add `!debug` to the message) which prints the parameters to each tool - should be useful to check if it's doing the right thing in production.

How it looks in normal and debug mode:

https://github.com/user-attachments/assets/ba0ead75-1407-4227-b03b-10f5e39b39b8

https://github.com/user-attachments/assets/1f2bbb64-1f45-47a5-a358-c172c0273b41

<details>
<summary><strong>Other example questions</strong></summary>

Screenshot 2025-04-12 at 21 07 07

---

Screenshot 2025-04-12 at 21 07 26

---

Screenshot 2025-04-12 at 21 12 43

---

Screenshot 2025-04-12 at 21 13 07

---

Screenshot 2025-04-12 at 21 13 41

---

Screenshot 2025-04-13 at 00 19 33

</details>

## After merging

Update Slack OAuth permissions, events, and bot online status setting to the same as my test app. Check through any other config in https://sdk.vercel.ai/docs/guides/slackbot#deploying-the-app


binary-koan (commented):

apps/nextjs/src/lib/data/conversation/search.ts
```diff
@@ -103,6 +106,22 @@ export const searchConversations = async (
           ),
         }
       : {}),
+  };
+
+  const matches = filters.search ? await searchEmailsByKeywords(filters.search, mailbox.id, Object.values(where)) : [];
```
binary-koan, line 111: Passing filters here massively speeds up searching open tickets in Gumroad since it can exclude closed ones, and makes results better since we won't filter out so many of the keyword search results

binary-koan:
It uses the same Slack integration we had before, just with more permissions. In general we need to document how to set integrations up (the README doesn't have proper instructions for the current Slack app setup either), I'll do some expanding/reworking of the docs this week.

raphaelcosta (approved):
@binary-koan _a Looks good! At some point, we should also store those conversations and their related events, so we can track them and figure out how to improve the agent.

binary-koan:
@raphaelcosta Ah yeah I was thinking that too. Are you aware of any good tooling for that or are you just imagining saving them in the database?

raphaelcosta:
Just saving in the database and we can use the OpenAI dashboard for any specific debugging.

PR #231, opened 2025-04-11
chore: generate Expo's tRPC types using tsc instead of xtrpc

binary-koan:
I've been having more and more problems with xtrpc not generating types correctly / erroring when the input and output types become nontrivial.

Saw this approach recommended in some other posts and it seems to work a lot better.

Blocker for Helper agent (it changes the search endpoint in a way that xtrpc refuses to parse).


apps/expo/.gitignore
```diff
@@ -24,3 +24,5 @@ expo-env.d.ts
 .env*.local
 
 eas.json
+
+src/trpc-types/
```
binary-koan, line 28: There's way too much generated code to be worth committing, so we need to run `expo:trpc:generate` locally now (it's already in `bin/dev` so I don't think it's an issue)

binary-koan (commented):

raphaelcosta (approved):
@binary-koan _a LGTM!

PR #241, opened 2025-04-16
chore: update README and use minio for local S3

binary-koan:
Adds instructions for setting up required integrations to the README, and adds Minio to docker-compose locally so you don't need to set up an S3 bucket for local development.


binary-koan:
@raphaelcosta _a Mind reviewing and checking that Minio works on your local? (works for me, but just want to check I didn't change some config outside Docker!)

Basically comment out the `vercel env pull` in `bin/dev`, comment out the AWS vars in `.env.local`, and file uploads should still work with files served from `minio.helperai.dev`

raphaelcosta (approved):
@binary-koan looks great! thanks

PR #250, opened 2025-04-18
chore: fix db reset script to delete enums

binary-koan:
@raphaelcosta FYI - the enums added for `guideSessionStatus` and `guideSessionEventType` weren't dropped when resetting the database so the reset script would run into "enum already exists" errors


raphaelcosta:
@binary-koan thanks!

PR #261, opened 2025-04-20
feat: send replies using the agent

binary-koan:
[ref](https://anti--work.slack.com/archives/C068HHZ64M9/p1745151598747579?thread_ts=1745074808.351449&cid=C068HHZ64M9)

Adds a two-step reply tool to the agent which first posts a message to confirm the text (allowing the user to edit it) and then creates the reply.

https://github.com/user-attachments/assets/abcf3e4b-3f26-44e8-b04e-4f96b4645826

### DIFF
--- a/apps/nextjs/src/lib/ai/index.ts
+++ b/apps/nextjs/src/lib/ai/index.ts
@@ -27,7 +27,12 @@ type CommonAIQueryOptions = {
   shortenPromptBy?: ShortenPromptOptions;
 };
 
-export const runAIQuery = async ({
+export const runAIQuery = async (options: Parameters<typeof runAIRawQuery>[0]): Promise<string> => {
+  const response = await runAIRawQuery(options);
+  return response.text;
+};
+
+export const runAIRawQuery = async ({
   messages,
   mailbox,
   queryType,
@@ -39,8 +44,8 @@ export const runAIQuery = async ({
   tools,
   functionId,
   shortenPromptBy,
-}: CommonAIQueryOptions & { maxSteps?: number; tools?: Record<string, Tool> }): Promise<string> => {
-  const response = await runWithRetry(
+}: CommonAIQueryOptions & { maxSteps?: number; tools?: Record<string, Tool> }) =>
+  await runWithRetry(
     () =>
       generateCompletion({
         messages,
@@ -58,8 +63,6 @@ export const runAIQuery = async ({
       }),
     { mailbox, queryType, model },
   );
-  return response.text;
-};
 
 export const runAIObjectQuery = async <T>({
   messages,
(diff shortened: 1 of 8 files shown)


apps/nextjs/src/lib/ai/index.ts
```diff
@@ -27,7 +27,12 @@ type CommonAIQueryOptions = {
   shortenPromptBy?: ShortenPromptOptions;
 };
 
-export const runAIQuery = async ({
+export const runAIQuery = async (options: Parameters<typeof runAIRawQuery>[0]): Promise<string> => {
+  const response = await runAIRawQuery(options);
+  return response.text;
+};
+
+export const runAIRawQuery = async ({
```
binary-koan, line 35: IMO we should make `runAIQuery` return the full result and update the other references; I'll do it separately if you agree

apps/nextjs/src/lib/slack/agent/getThreadMessages.ts
```diff
@@ -156,14 +167,10 @@ export const getThreadMessages = cache(
...
-      let content = "";
-
-      if (message.blocks && message.blocks.length > 0) {
-        content = convertKnownBlocksToMarkdown(message.blocks as Block[]);
-      }
+      let content = messageToMarkdown({ text: message.text, blocks: message.blocks as Block[] });
 
-      if (!content && message.text) {
-        content = message.text;
+      if (message.attachments && message.attachments.length > 0) {
```
binary-koan, line 172: Needed for VIP messages - we send the content as attachments instead of blocks directly

binary-koan (commented):

apps/nextjs/src/lib/slack/agent/generateAgentResponse.ts
```diff
@@ -52,226 +326,23 @@ IMPORTANT GUIDELINES:
 - Never share sensitive information or personal data
 - Don't discuss your own capabilities, programming, or AI nature unless directly relevant to answering the question
 - When listing tickets, display the standardSlackFormat field as is. You may add other information after that if relevant in context.
+- If you will need to reply to a ticket as part of your response and the sendReply tool is not available, use the confirmReplyText tool and *do not do anything else* at this stage.
+- *If you have the sendReply tool, call it!* Then include in your response that the reply has been sent.
```
raphaelcosta, line 330: You can force the tool usage by using the `toolChoice` parameter https://sdk.vercel.ai/docs/ai-sdk-core/tools-and-tool-calling#tool-choice

raphaelcosta (approved):
@binary-koan _a LGTM!

apps/nextjs/src/lib/slack/agent/generateAgentResponse.ts
binary-koan: Just tried that and it sent the reply over and over 🤔 Will merge as is and look into that option more if the prompt isn't enough

PR #309, opened 2025-04-27
feat: integrate docs into main app

binary-koan:


binary-koan:
Still trying to work out why autofix isn't happy, but will merge for now so we have a non-broken docs site

PR #317, opened 2025-04-29
fix: responsive widget for mobile

raphaelcosta:
- Use the full-page widget for the mobile layout, just like we were doing before with the drawer

Screenshot 2025-04-29 at 8 43 59 AM

### DIFF
--- a/packages/nextjs/sdk/embed.css
+++ b/packages/nextjs/sdk/embed.css
@@ -368,6 +368,10 @@
     bottom: 16px;
     padding-bottom: env(safe-area-inset-bottom, 16px);
   }
+  
+  .helper-widget-wrapper.temporarily-hidden {
+    opacity: 0;
+  }
 }
 
 @keyframes slide-in {
--- a/packages/nextjs/sdk/embed.css
+++ b/packages/nextjs/sdk/embed.css
@@ -154,6 +154,15 @@
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/nextjs/sdk/embed.css
```diff
@@ -287,15 +296,82 @@
...
+    bottom: 0;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    padding-top: env(safe-area-inset-top, 0px);
+    padding-bottom: env(safe-area-inset-bottom, 0px);
+    padding-left: env(safe-area-inset-left, 0px);
+    padding-right: env(safe-area-inset-right, 0px);
+    border-radius: 0;
```
slavingia, line 310:
```suggestion
    border-radius: 4px;
```

packages/nextjs/sdk/embed.css
```diff
@@ -287,15 +296,82 @@
...
+    box-shadow: none;
```
slavingia, line 311: IMO we should keep, but make it softer generally.

slavingia (approved):

packages/nextjs/sdk/embed.css
raphaelcosta: It won’t be visible because the widget takes the entire page

PR #344, opened 2025-04-30
feat: switch to pnpm

binary-koan:
* Sensible lockfile handling and dependency upgrades
* Sensible handling of arguments in scripts
* Better at running commands in workspaces

Also upgraded Node to 22


.npmrc
```diff
@@ -1,2 +1 @@
-node-linker=hoisted
-link-workspace-packages=true
+shamefully-hoist=true
```
binary-koan, line 1: Needed to fix the require-in-the-middle issue described here: https://docs.sentry.io/platforms/javascript/troubleshooting/#pnpm-resolving-import-in-the-middle-external-package-errors

PR #394, opened 2025-05-07
feat: docker-only dev setup

binary-koan:
(Still experimental)


binary-koan:
I like this idea but I think we'll want people to have Node installed anyway for scripts. Shelving for now

PR #401, opened 2025-05-08
Change file path arg for seedDatabase.ts

mattbrc:
Context: ran into `Database reset failed: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined` on indexed message 37 when running `pnpm db:reset`

Issue: `import.meta.dirname` evaluated to undefined resulting in typeerror on path.join

Resolution: swap `import.meta.dirname` for `dirname(fileURLToPath(import.meta.url))`

### DIFF
--- a/db/seeds/seedDatabase.ts
+++ b/db/seeds/seedDatabase.ts
@@ -152,7 +152,7 @@ export const seedDatabase = async () => {
     }
 
     // Optionally create this file to do any additional seeding, e.g. setting up integrations with local credentials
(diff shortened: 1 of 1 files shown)


binary-koan:
What Node version are you on? If you follow `.node-version` it should be supported https://nodejs.org/api/esm.html#importmetadirname

mattbrc:
Ahh ok that was it, thanks @binary-koan !

PR #406, opened 2025-05-10
Invalidate preferences fetch on save

mattbrc:
Issue: On making a change in preferences (mailbox name, confetti, theme) and saving, next cache does not force a refetch when navigating to another tab and back to preferences unless a hard refresh is done. 

Resolution: invalidate mailbox fetch on saved changes in preferences

## Summary by CodeRabbit

- **Bug Fixes**
	- Ensured mailbox settings and preferences are refreshed immediately after updates, providing up-to-date information in the dashboard.


binary-koan (approved):
Thanks for the fix! I'm about to rework a bunch of those settings queries so it's nice to have one less thing to update!

Issue #404, opened 2025-05-10
Improve search performance


binary-koan:
Unsure of the best option, but some ideas:
* Use Postgres better - better indexing, smarter queries
* Use a third party like [Meilisearch](https://www.meilisearch.com) or Elasticsearch (if so, this should ideally be optional to simplify setup for smaller sites)
* [turbopuffer](https://x.com/turbopuffer/status/1810686962707087526?s=46)

Also we should index internal notes as well as messages so staff can note search terms.

### Why

Searching through closed tickets gets extremely slow as the number of tickets increases. The relevance also kinda sucks, it just lists the newest tickets that contain at least one of the search keywords.

MJomaa:
Question is how do you rank top-k relevant elements efficiently.
 
- Elasticsearch and Solr both use Lucene under the hood, which defaults to the BM25 similarity algorithm for ranking search results. BM25 provides high-quality relevance ranking, but it requires  pre-processing like analyzers, token filters, custom mappings and is not lightweight in terms of compute/storage.

- PostgreSQL Full-Text Search (FTS) uses a TF-IDF-inspired algorithm (with coverage density ranking via `ts_rank_cd`) that provides BM25-like quality, just slightly worse but it comes close enough. It is simpler to set up, requires less pre-processing and is more resource-efficient.
 
 Imo just use FTS:
- Create and store the ts_vector in an extra column.
- Create a GIN index on that column.
- Optionally create some weights
- When selecting, use ts_rank_cd  and convert your query to the tsquery data type.

Are the messages very similar in nature or do you think each one contains at least one less frequent word? If they are super similar we might need a different ranking.

slavingia:
Created a turbopuffer account, seems like a good solution personally.

slavingia:
I guess it can create lock-in, so maybe Supabase is better but didn't want to be a bottleneck to improving our own UX.

binary-koan:
Thanks, will look into it. We can make things like that optional IMO and fall back to basic db search if they're not set up

daniellockyer:
I have some questions:

1. the issue is written assuming the bottleneck is the database. How do we know this is the case? Do we have instrumentation showing this? (for example, I've seen performance issues that look like the DB but it was actually the in-app serialization/deserialization)
1. What is the best way to reproduce this locally?

binary-koan:
[...]

PR #426, opened 2025-05-19
perf: improved search performance by caching key derivation

daniellockyer:
refs https://github.com/antiwork/helper/issues/404

- when searching, the slowest part of the flow is decrypting the `cleanedUpText`, which is encrypted using a lib we have locally
- instead of deriving the key for every decryption (which happens a lot during searching), we can just calculate this once
- this makes searching 10x faster on my local dev env

## Summary by CodeRabbit

- **Refactor**
	- Improved performance by caching encryption keys and reusing cryptographic instances internally, resulting in faster encryption and decryption operations. No changes to user-facing interfaces.


binary-koan (approved):
Nice find!

PR #429, opened 2025-05-19
fix: immediately remove theme when disabled

binary-koan:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved theme settings so that theme colors reset correctly whenever the theme is disabled.


binary-koan:
Doesn't work, it's a bit of a pain to make it work with how next-themes does class names. Will see if we can remove next-themes at some point

PR #445, opened 2025-05-22
fix: always reuse the thread mailbox in agent

binary-koan:
[ref](https://anti--work.slack.com/archives/C055QQCQFM1/p1747816821798259)

### DIFF
--- a/lib/slack/agent/generateAgentResponse.ts
+++ b/lib/slack/agent/generateAgentResponse.ts
@@ -385,7 +385,7 @@ export const generateAgentResponse = async (
     model: "gpt-4o",
     system: `You are Helper's Slack bot assistant for customer support teams. Keep your responses concise and to the point.
 
-You are currently in the mailbox: ${mailbox.name}.
+You are currently in the mailbox: ${mailbox.name}. You cannot access any other mailboxes; the user must start a new chat to access others.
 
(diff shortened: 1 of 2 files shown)


lib/slack/agent/generateAgentResponse.ts
```diff
@@ -385,7 +385,7 @@ export const generateAgentResponse = async (
     model: "gpt-4o",
     system: `You are Helper's Slack bot assistant for customer support teams. Keep your responses concise and to the point.
 
-You are currently in the mailbox: ${mailbox.name}.
+You are currently in the mailbox: ${mailbox.name}. You cannot access any other mailboxes; the user must start a new chat to access others.
```
slavingia, line 388:
```suggestion
You are currently in the mailbox: ${mailbox.name}. You cannot access any other mailboxes; the user must start a new chat or explicitly mention another mailbox by name to access others.
```

slavingia: As sometimes I talk to helper via DM

binary-koan: I imagined starting a new chat in DMs each time, but you're right it would be nice to just switch on the fly. Will do separately as I'll need to fix the actual mailbox detection

PR #450, opened 2025-05-27
perf: added DB index for `conversations_conversation.merged_into_id`

daniellockyer:
- as identified in Supabase, this is used in one of the slowest and most time consuming queries we have
- adding this index reduces the cost by an estimated 99.95%

image


binary-koan (approved):
LGTM! (same tests are failing on main, will fix)

PR #457, opened 2025-05-28
perf: parallelised inngest sending for multiple events

daniellockyer:
- instead of running them in serial, we can run them in parallel


daniellockyer:
Checking tests

Issue #490, opened 2025-06-05
Improve theming support


binary-koan:
- [ ] Move theme colors to globals.css and remove the database setting
- [ ] Make the site more themeable - try some examples and fix places where foreground/background colours don't match
- [ ] Update the widget to be fully themeable (ideally themed separately to the mailbox)
- [ ] Might be nice to have a preview UI which gives you a snippet to copy, might be overkill
- [ ] Add a doc

### Why

Make Helper more customizable - people trying it out often bring up theming.

slavingia:
Will be done by moving to api

PR #501, opened 2025-06-11
feat: add reset button for theme color inputs

Adarsh9977:
### Add reset buttons to reset theme colours
need some suggestions on this :- 
- Instead of resetting one by one we can also create a single reset button to reset the theme to default.
- Instead of multiple inputs for different theme properties we can make single component and pass the theme, onChange, onClick for optimal and clean code
 
- After Changes

https://github.com/user-attachments/assets/4cb35a0a-8ae2-48d1-971b-8e89fe0c0f8d

 

## Summary by CodeRabbit

- **New Features**
  - Added reset buttons next to each theme color input, allowing users to quickly revert colors to their default values.

- **Style**
  - Improved layout of theme color inputs by grouping each input with its corresponding reset button for a cleaner interface.


binary-koan:
We're actually planning to remove this UI and just improve theme support via CSS edits: https://github.com/antiwork/helper/issues/490

This would otherwise be a great feature, appreciate the thought!

PR #552, opened 2025-06-21
feature: add in conversation search for all the conversations

sankalp1999:
### Smart Message Navigation

- Shows "X of Y matches" (messageItems in which word is found) with dedicated next/previous navigation
- Auto-scrolls and centers each message containing search results
- Keyboard shortcuts: Enter (next), Shift+Enter (previous), Escape (close)

### Conversation-Scoped search

-Searches only within conversation messages and sender names, 
- Ignores sidebar, navigation, and UI elements that clutter browser search results. 
- Focuses user attention on relevant conversation content

## Why It's better than browser search?

Browser search treats conversations like static web pages - it highlights random UI text, navigation elements, and buttons alongside actual conversation content. This creates noise and makes it hard to find what you're actually looking for.
The in-conversation search understands the conversation structure - it knows what's a message, who sent it, and how to navigate between relevant results. Instead of 47 random highlights across buttons and menus, you get "3 of 8 matches" in actual conversation content with smart navigation between them.

Bottomline the user can quickly find relevant text without getting distracted by irrelevant page elements or losing context while jumping between matches.

## Video showing the feature

https://github.com/user-attachments/assets/062d6525-e119-47ab-9346-75f95685c0f5

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx
@@ -3,17 +3,20 @@ import {
   ArrowLeft,
   ArrowRight,
   ArrowUp,
+  ChevronDown,
   ChevronLeft,
   ChevronRight,
+  ChevronUp,
   Download,
   Info,
   Link as LinkIcon,
   PanelRightClose,
   PanelRightOpen,
+  Search,
   X,
 } from "lucide-react";
 import Link from "next/link";
-import { useEffect, useLayoutEffect, useState } from "react";
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from "react";
 import { useMediaQuery } from "react-responsive";
 import { useStickToBottom } from "use-stick-to-bottom";
 import { create } from "zustand";
@@ -40,6 +43,7 @@ import { Alert, AlertDescription, AlertTitle } from "@/components/ui/alert";
 import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
 import { Dialog, DialogContent, DialogHeader, DialogTitle } from "@/components/ui/dialog";
+import { Input } from "@/components/ui/input";
 import { ResizableHandle, ResizablePanel, ResizablePanelGroup } from "@/components/ui/resizable";
 import { SidebarProvider } from "@/components/ui/sidebar";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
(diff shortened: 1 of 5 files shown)


ConstantTime:
I think there is one issue with this PR. This introduces client-side plaintext search while the cross-conversation search in helper currently (in the `mailboxes` page) uses hash-based privacy protection. Checkout this: https://github.com/antiwork/helper/blob/main/lib/emailSearchService/extractHashedWordsFromEmail.ts

sankalp1999:
thanks, will have a look

sankalp1999:
The in-conversation search operates on data that was already decrypted on server for display purposes so I think this is fine.  I am not doing any additional decryption. Just searching through content that was already decrypted. 

It's basically operating on data that has already been decrypted for user's viewing session.


Some reasoning behind this

In lib/data/conversationMessage.ts:210-215, the [serializeMessage](https://github.com/antiwork/helper/blob/147547ccf4fa3acece8bf90c0fb9b08eb482d9d8/lib/data/conversationMessage.ts#L188) function decrypts message bodies before sending them to the client for display.

At [line 143](https://github.com/antiwork/helper/blob/147547ccf4fa3acece8bf90c0fb9b08eb482d9d8/lib/data/conversationMessage.ts#L146) in same file, all messages are going through `serializeMessage` function so they are decrypted server side.

In `app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx:350`, the search function accesses message.body which contains the already-decrypted content that was sent for display.

sankalp1999:
The privacy boundary is same as earlier for in convo search -  "authorized conversation viewing", once a user opens a conversation, all content is decrypted for their session.

app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx
```diff
@@ -57,6 +62,84 @@ export type ConversationWithNewMessages = Omit<ConversationType, "messages"> & {
...
 };
 
+export type SearchState = {
+  query: string;
+  isActive: boolean;
+  matches: { messageId: string; messageIndex: number; matchIndex: number }[];
+  currentMatchIndex: number;
+};
+
+export const useConversationSearchStore = create<{
```
Comment (binary-koan) [inline app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx:72]: This file is already huge, would be nice to put this in a separate file

app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx
```diff
@@ -231,13 +337,119 @@ const ConversationHeader = ({
...
   return (
     <div
       className={cn(
-        "flex items-center border-b border-border h-12 px-2 md:px-4 gap-x-2",
+        "flex items-center border-b border-border px-2 md:px-4 gap-x-2",
         !conversationInfo && "hidden",
+        searchState.isActive ? "h-auto min-h-12 py-2" : "h-12",
       )}
       style={{ minHeight: 48 }}
+      data-conversation-area
```
Comment (binary-koan) [inline app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx:452]: Why is this on the header as well?

binary-koan (commented):
I really like this overall! Added a few comments; also would be easier to navigate if there's some kind of "pop" animation on each result as you go through them (like Safari's in-page search for example)

binary-koan:
Hmm yeah highlighting in HTML sounds tricky. Will look into that when I next review

sankalp1999:
@binary-koan I think this is complete. I am able to highlight text within html also. actually fixed a bug which led to html rendering in our messages too.

> ## Other bug fixes
> ### Before - some email messages are pure html elements and they were being rendered as such
> Screenshot 2025-06-24 at 2 25 37 PM
> ### After - now they render properly and we can search through these messages
> Screenshot 2025-06-24 at 2 30 42 PM

sankalp1999:
now CI should pass

sankalp1999:
>Bug: Role Display Bypassed in Active Search
intentional, i want to limit search to the messages

sankalp1999:
@slavingia if i understood correctly, you mean that we can keep search bar open like this and show the conversation title in browser's title tag right?


Screenshot 2025-06-24 at 9 11 27 PM

binary-koan:
The title is [supposed to be set already](https://github.com/antiwork/helper/blob/14ce8637bb3890c019cfce81d9dad7d9a3fc28c6/app/(dashboard)/mailboxes/%5Bmailbox_slug%5D/%5Bcategory%5D/inbox.tsx#L103) - I think the Next.js and React 19 ways are conflicting :(

Maybe you can update that place to use `useEffect` instead of adding a new one?

sankalp1999:
removed useeffect i added recently and updated in that one in latest change @binary-koan 

https://github.com/user-attachments/assets/d9884472-9b33-47e1-991c-76b754aa7bae

slavingia:
The change to add titles generally would be good though!

sankalp1999:
Ok i will push that.

PR #560, opened 2025-06-22
enhancement: move the toolbar for formatting such that it doesnt cover the cursor

debadree25:
This is a small tweak to how the formatting toolbar appears when typing:

Before (it covers the cursor and the typed content):

https://github.com/user-attachments/assets/471f84cd-e20d-496d-a7ab-de75564a6eff

After (small amount of spacing added when toolbar is expanded):

https://github.com/user-attachments/assets/843ffe3f-98a1-4e12-8b86-5e3b361b1587

### DIFF
--- a/components/tiptap/toolbar.tsx
+++ b/components/tiptap/toolbar.tsx
@@ -308,7 +308,7 @@ const Toolbar = ({
       <div
         className={cn(
           isAboveMd
-            ? "flex flex-wrap gap-1 absolute z-10 bottom-16 mb-2 right-3 rounded-t border rounded-sm bg-background p-1"
+            ? `flex flex-wrap gap-1 absolute z-10 ${open ? "bottom-11" : "bottom-16"} mb-2 right-3 rounded-t border rounded-sm bg-background p-1`
             : "flex flex-1 min-w-0 gap-1",
           open && isAboveMd && "left-3",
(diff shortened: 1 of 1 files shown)


components/tiptap/toolbar.tsx
```diff
@@ -308,7 +308,7 @@ const Toolbar = ({
       <div
         className={cn(
           isAboveMd
-            ? "flex flex-wrap gap-1 absolute z-10 bottom-16 mb-2 right-3 rounded-t border rounded-sm bg-background p-1"
+            ? `flex flex-wrap gap-1 absolute z-10 ${open ? "bottom-11" : "bottom-16"} mb-2 right-3 rounded-t border rounded-sm bg-background p-1`
```
binary-koan, line 311: This will break the reply input, won't it? I think the issue is that this needs to be different for reply vs new conversation (should always be bottom aligned on both)

debadree25: yes indeed it does i am thinking would controlling this behaviour by passing props would be a good idea?

debadree25: Understood updating accordingly

binary-koan:
Closing as we won't merge this in its current state - feel free to reopen or open a new PR with the suggested changes

PR #563, opened 2025-06-22
perf: hoist members query to reduce unnecessary re-renders

sankalp1999:
- Move api.organization.getMembers query from AssignedToLabel to parent ConversationList component
- Pass members data down as props to ConversationListItem components
- Eliminates redundant useQuery hooks that caused unnecessary re-renders
- Each AssignedToLabel no longer manages its own query state

## Summary by CodeRabbit

- **Refactor**
  - Improved performance and consistency by fetching organization member data once and sharing it across conversation list items.
  - Streamlined display of user assignment by passing member data directly to relevant components.
  - Introduced a context provider to supply organization member data throughout the mailbox layout for better state management.

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/assignPopoverButton.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/assignPopoverButton.tsx
@@ -2,6 +2,7 @@
 
 import { Bot, User } from "lucide-react";
 import { useEffect, useState } from "react";
+import { useOrganizationMembers } from "@/app/(dashboard)/mailboxes/[mailbox_slug]/organizationMembersContext";
 import { AssigneeOption, AssignSelect } from "@/components/assignSelect";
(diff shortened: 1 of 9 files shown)


binary-koan:
I don't quite understand the purpose of this - won't the number of renders be the same because it'll still re-render the child components if the query result changes?

The point of `staleTime: Infinity` etc in this query is to avoid this extra indirection, contexts, etc, by making sure the same data is used everywhere and preventing refetches (which AFAIK should make the overhead of useQuery minimal)

sankalp1999:
@binary-koan What I was trying to do with this PR was reduce the number of api.organization.getMembers.useQuery calls happening - each assigneeFilter, responderFilter would do a call. Although React Query deduplicates network requests, so there's no real performance difference.

Despite above, I was thinking that maybe we could fetch data at the top level, but realized React Query already solves for this with its global cache (I think your comment hints towards this only). This PR is just increasing complexity without much benefit. Thanks for the review - I'll close it.

PR #564, opened 2025-06-22
feat: remove member from the team

Adarsh9977:
## User can remove members from the team

- Before

Screenshot 2025-06-23 at 4 02 11 AM

- After

https://github.com/user-attachments/assets/c5872595-1496-48db-898e-9ab145a4a67e

##  doubts

- can a single member remove himself from the team.

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx
@@ -9,6 +9,8 @@ import { TableCell, TableRow } from "@/components/ui/table";
 import { useDebouncedCallback } from "@/components/useDebouncedCallback";
 import { type UserRole } from "@/lib/data/user";
 import { api } from "@/trpc/react";
+import { Button } from "@/components/ui/button";
+import { Trash } from "lucide-react";
 
 export const ROLE_DISPLAY_NAMES: Record<UserRole, string> = {
   core: "Core",
@@ -27,9 +29,10 @@ interface TeamMember {
 type TeamMemberRowProps = {
   member: TeamMember;
   mailboxSlug: string;
+  onDelete: (id: string, name: string)=> void;
 };
 
-const TeamMemberRow = ({ member, mailboxSlug }: TeamMemberRowProps) => {
+const TeamMemberRow = ({ member, mailboxSlug, onDelete }: TeamMemberRowProps) => {
   const [keywordsInput, setKeywordsInput] = useState(member.keywords.join(", "));
   const [role, setRole] = useState<UserRole>(member.role);
   const [localKeywords, setLocalKeywords] = useState<string[]>(member.keywords);
@@ -172,6 +175,20 @@ const TeamMemberRow = ({ member, mailboxSlug }: TeamMemberRowProps) => {
           placeholder="Enter keywords separated by commas"
           className={role === "nonCore" ? "" : "invisible"}
         />
+      </TableCell>
+      <TableCell>
+        <Button
+          variant="ghost"
+          size="sm"
+          iconOnly
+          onClick={(e) => {
+            e.preventDefault();
+            onDelete(member.id, member.displayName);
+          }}
+        >
+          <Trash className="h-4 w-4" />
+          <span className="sr-only">Delete</span>
+        </Button>
       </TableCell>
     </TableRow>
   );
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx
@@ -1,16 +1,16 @@
 "use client";
 
+import { Trash } from "lucide-react";
 import { useEffect, useState } from "react";
 import { toast } from "@/components/hooks/use-toast";
 import { Avatar } from "@/components/ui/avatar";
+import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "@/components/ui/select";
 import { TableCell, TableRow } from "@/components/ui/table";
 import { useDebouncedCallback } from "@/components/useDebouncedCallback";
 import { type UserRole } from "@/lib/data/user";
 import { api } from "@/trpc/react";
-import { Button } from "@/components/ui/button";
-import { Trash } from "lucide-react";
 
 export const ROLE_DISPLAY_NAMES: Record<UserRole, string> = {
   core: "Core",
@@ -29,7 +29,7 @@ interface TeamMember {
 type TeamMemberRowProps = {
   member: TeamMember;
   mailboxSlug: string;
-  onDelete: (id: string, name: string)=> void;
+  onDelete: (id: string, name: string) => void;
 };
 
 const TeamMemberRow = ({ member, mailboxSlug, onDelete }: TeamMemberRowProps) => {
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx
@@ -2,6 +2,7 @@
 
 import { Search } from "lucide-react";
 import { useState } from "react";
+import { toast } from "@/components/hooks/use-toast";
 import LoadingSpinner from "@/components/loadingSpinner";
 import { Input } from "@/components/ui/input";
 import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from "@/components/ui/table";
@@ -9,7 +10,6 @@ import { api } from "@/trpc/react";
 import SectionWrapper from "../sectionWrapper";
 import { AddMember } from "./addMember";
 import TeamMemberRow, { ROLE_DISPLAY_NAMES } from "./teamMemberRow";
-import { toast } from "@/components/hooks/use-toast";
 
 type TeamSettingProps = {
   mailboxSlug: string;
@@ -47,10 +47,10 @@ const TeamSetting = ({ mailboxSlug }: TeamSettingProps) => {
   });
 
   const removeMember = (id: string, name: string) => {
-    if(confirm(`Are you sure you want to remove ${name} from your team?`)){
+    if (confirm(`Are you sure you want to remove ${name} from your team?`)) {
       removeMemberMutation({ id });
     }
-  }
+  };
 
   return (
     <SectionWrapper
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx
@@ -9,6 +9,7 @@ import { api } from "@/trpc/react";
 import SectionWrapper from "../sectionWrapper";
 import { AddMember } from "./addMember";
 import TeamMemberRow, { ROLE_DISPLAY_NAMES } from "./teamMemberRow";
+import { toast } from "@/components/hooks/use-toast";
 
 type TeamSettingProps = {
   mailboxSlug: string;
@@ -16,6 +17,24 @@ type TeamSettingProps = {
 
 const TeamSetting = ({ mailboxSlug }: TeamSettingProps) => {
   const { data: teamMembers = [], isLoading } = api.mailbox.members.list.useQuery({ mailboxSlug });
+  const utils = api.useUtils();
+  const { mutate: removeMemberMutation, isPending: isRemoving } = api.organization.removeMember.useMutation({
+    onSuccess: () => {
+      toast({
+        title: "Team member removed",
+        variant: "success",
+      });
+
+      utils.mailbox.members.list.invalidate({ mailboxSlug });
+    },
+    onError: (error) => {
+      toast({
+        title: "Failed to send invitation",
+        description: error.message,
+        variant: "destructive",
+      });
+    },
+  });
   const [searchTerm, setSearchTerm] = useState("");
 
   const filteredTeamMembers = teamMembers.filter((member) => {
@@ -26,6 +45,12 @@ const TeamSetting = ({ mailboxSlug }: TeamSettingProps) => {
       member.keywords.some((keyword) => keyword.toLowerCase().includes(searchString))
     );
   });
+
+  const removeMember = (id: string, name: string) => {
+    if(confirm(`Are you sure you want to remove ${name} from your team?`)){
+      removeMemberMutation({ id });
+    }
+  }
 
   return (
     <SectionWrapper
@@ -74,7 +99,7 @@ const TeamSetting = ({ mailboxSlug }: TeamSettingProps) => {
                 </TableRow>
               ) : (
                 filteredTeamMembers.map((member) => (
-                  <TeamMemberRow key={member.id} member={member} mailboxSlug={mailboxSlug} />
+                  <TeamMemberRow key={member.id} member={member} mailboxSlug={mailboxSlug} onDelete={removeMember} />
                 ))
               )}
             </TableBody>
--- a/lib/data/user.ts
+++ b/lib/data/user.ts
@@ -6,18 +6,6 @@ import { getFullName } from "@/lib/auth/authUtils";
 import { createAdminClient } from "@/lib/supabase/server";
 import { getSlackUser } from "../slack/client";
 
-export const addUser = async (inviterUserId: string, emailAddress: string, displayName: string) => {
-  const supabase = createAdminClient();
-  const { error } = await supabase.auth.admin.createUser({
-    email: emailAddress,
-    user_metadata: {
-      inviter_user_id: inviterUserId,
-      display_name: displayName,
-    },
-  });
-  if (error) throw error;
-};
-
 export const UserRoles = {
   CORE: "core",
   NON_CORE: "nonCore",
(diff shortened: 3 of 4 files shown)


lib/data/user.ts
```diff
@@ -40,6 +28,26 @@ export type UserWithMailboxAccessData = {
...
+  if (error) throw error;
+};
+
+export const removeUser = async (id: string) => {
+  const supabase = createAdminClient();
+
+  const { error } = await supabase.auth.admin.deleteUser(id);
+
+  if (error) throw error;
+};
```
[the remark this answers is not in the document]
Adarsh9977: protectedProcedure in trpc works as a middleware to check the authorization this function is the last call for deletion.

app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx
```diff
@@ -74,7 +100,7 @@ const TeamSetting = ({ mailboxSlug }: TeamSettingProps) => {
                 </TableRow>
               ) : (
                 filteredTeamMembers.map((member) => (
-                  <TeamMemberRow key={member.id} member={member} mailboxSlug={mailboxSlug} />
```
Comment (slavingia) [inline app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx:77]: Use shadcn dialog not native confirm

slavingia (changes requested):

app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx
```diff
@@ -27,9 +29,10 @@ interface TeamMember {
 type TeamMemberRowProps = {
   member: TeamMember;
   mailboxSlug: string;
+  onDelete: (id: string, name: string) => void;
```
Comment (slavingia) [inline app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx:32]: Do we need this? Didn't for adding members.

And only admins should be able to remove and can't remove self

Comment (Adarsh9977) [reply app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx:32]: @slavingia Thanks! also a follow up feature if you confirm will raise a pr for that tomorrow. Instead of editable name and role as default we can add an edit icon in actions and when clicks edit button then only make that specific row editable.

Screenshot 2025-06-23 at 10 34 21 PM

app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx
```diff
@@ -173,6 +194,17 @@ const TeamMemberRow = ({ member, mailboxSlug }: TeamMemberRowProps) => {
           className={role === "nonCore" ? "" : "invisible"}
         />
       </TableCell>
+      <TableCell>
+        <ConfirmationDialog
+          message={`Are you sure you want to remove ${member.displayName} from your team?`}
+          onConfirm={() => removeTeamMember({ id: member.id, mailboxSlug })}
+        >
+          <Button variant="ghost" size="sm" iconOnly>
```
Comment (binary-koan) [inline app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamMemberRow.tsx:202]: We should hide this for the current user

lib/data/user.ts
```diff
@@ -40,6 +28,26 @@ export type UserWithMailboxAccessData = {
...
+      display_name: displayName,
+    },
+  });
+  if (error) throw error;
+};
+
+export const removeUser = async (id: string) => {
+  const supabase = createAdminClient();
+
+  const { error } = await supabase.auth.admin.deleteUser(id);
```
binary-koan, line 46: We'll need to remove their ticket assignments and check through other tables for anything that might break

Or ideally we'd disable rather than deleting I think so that historical events still show the user name correctly (maybe ban for a super long time is the Supabase way to do it? https://github.com/orgs/supabase/discussions/9239)

binary-koan (commented):
Thanks! Added a couple of comments

lib/data/user.ts
slavingia: We should mark users as deleted_at instead of hard deleting and then we don't have to do this extra cleanup IMO. Though it may be good to have this in a dialog with conforming a team member removal how many tickets they're assigned to (if > 0) and a select to assign them to someone else.

Adarsh9977: @slavingia Thanks! Also another approach we can remove the mailboxAccess of that mailbox for that user instead of deleting or mark as deleted.

Adarsh9977:
@binary-koan Please verify the new approach 
```
export const getUsersWithMailboxAccess = async (mailboxId: number): Promise<UserWithMailboxAccessData[]> => {
  const users = await db.query.authUsers.findMany();

  return users
    .filter((user) => {
      const mailboxAccess = (user.user_metadata?.mailboxAccess ?? {}) as Record<string, any>;
      return mailboxAccess[mailboxId];
    })
    .map((user) => {
      const mailboxAccess = (user.user_metadata?.mailboxAccess ?? {}) as Record<string, any>;
      const access = mailboxAccess[mailboxId];

      return {
        id: user.id,
        displayName: user.user_metadata?.display_name || "",
        email: user.email ?? undefined,
        role: access?.role || "afk",
        keywords: access?.keywords || [],
      };
    });
};
```
 and 

``` 
export const removeMailboxAccess = async (id: string, mailboxId: number) => {
  const user = await db.query.authUsers.findFirst({
    where: eq(authUsers.id, id),
  });

  if (!user) throw new Error("User not found");

  const mailboxAccess = (user.user_metadata?.mailboxAccess ?? {}) as Record<string, any>;

  delete mailboxAccess[mailboxId];

  await db
    .update(authUsers)
    .set({
      user_metadata: {
        ...user.user_metadata,
        mailboxAccess,
      },
    })
    .where(eq(authUsers.id, id));
};
````

Also we have to update our AddMember logic too for adding a new member with 

```
export const addUser = async (
  inviterUserId: string,
  emailAddress: string,
  displayName: string,
  mailboxId: number,
  role: "core" | "non-core" | "away" = "away" // default role if not provided
) => {
  const supabase = createAdminClient();

  const { error } = await supabase.auth.admin.createUser({
    email: emailAddress,
    user_metadata: {
      inviter_user_id: inviterUserId,
      display_name: displayName,
      mailboxAccess: {
        [mailboxId]: {
          role: role,
          keywords: [],
          updatedAt: new Date().toISOString(),
        },
      },
    },
  });

  if (error) throw error;
};
```

This approach helps us to maintain a better connection with only accessed members

**Comment by binary-koan** [comment]
I would rather use bans for two reasons

- `user_metadata` can be updated on the frontend, so if the user can still log in with the Supabase client then they can just add themselves back again
- We're removing multiple mailbox support, so there isn't a case where a user will have access to one mailbox and not another

wdyt?

**Comment by Adarsh9977** [comment]
@binary-koan Implemented the ban functionality with a ban limit of 50 years, but that user will not be able to login again into our website until we add back or removed his ban. Is this the right format. If you are ok will push that.

**Comment by slavingia** [comment]
May be good to merge in admin role then this one and only admin can add/remoce

**Comment by Adarsh9977** [comment]
> May be good to merge in admin role then this one and only admin can add/remoce

Yes I created a pr for admin role https://github.com/antiwork/helper/pull/580.

**Comment by binary-koan** [comment]
I'd prefer if we kept the rows in the dialog after you select someone (in case of mistakes), and instead added an explicit button to proceed that gets enabled once everything is reassigned. In that case it could replace the confirmation dialog actually, we could make the button text "Reassign and delete member" or something

**Comment by slavingia** [comment]
We can just say the number of tickets with a select for all; don't need to be specific per ticket for this release IMO.

image

**Comment by Adarsh9977** [comment]
> We can just say the number of tickets with a select for all; don't need to be specific per ticket for this release IMO.
> 
> image

Ok, will push the new changes in a while. btw I was also thinking about the same approach

**Comment by Adarsh9977** [comment]
```
-- Function to automatically insert into user_profiles after a user is created
create function public.handle_new_user_profile()
returns trigger
language plpgsql
security definer set search_path = ''
as $$
begin
  insert into public.user_profiles (id, display_name, permissions, access, created_at, updated_at)
values (
  new.id,
  new.raw_user_meta_data ->> 'display_name',
  'member',                         -- default permissions
  jsonb_build_object('role', 'afk', 'keywords', '[]'), -- default access
  now(),                             -- default created_at
  now()                              -- default updated_at
);
  return new;
end;
$$;

-- Trigger to call the function after insert on auth.users
create trigger on_auth_user_created
after insert on auth.users
for each row
execute procedure public.handle_new_user_profile();
```

@binary-koan by some research and through I created this trigger that helps us to maintain a public table whenever user signUp we save default fields inside the public table.After signup, our API only interacts with the public user_profiles table.
right now i had created specific columns for access(role), permissions. In future we'll only interacts with them. Please take a look and review/guide.

**Comment by binary-koan** [comment]
Marking this as draft until https://github.com/antiwork/helper/pull/580 is merged

File: app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx [hunk 6]
```diff
@@ -30,6 +32,23 @@ const TeamSetting = ({ mailboxSlug }: TeamSettingProps) => {
...
+      utils.mailbox.members.list.invalidate({ mailboxSlug });
+    },
+    onError: (error) => {
+      toast({
+        title: "Failed to remove member",
+        description: error.message,
+        variant: "destructive",
+      });
+    },
+  });
```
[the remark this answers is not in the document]
Comment (Adarsh9977) [reply app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx]: we are doing ```utils.mailbox.members.list.invalidate({ mailboxSlug });``` for invalidate the conversations list on teamSettings page

Adarsh9977:
## when user is admin

Screenshot 2025-07-05 at 2 52 04 AM

## When used is not admin/member
Screenshot 2025-07-05 at 2 56 52 AM

## when no assigned conversations
Screenshot 2025-07-05 at 3 18 54 AM

## when assigned conversations
Screenshot 2025-07-05 at 3 19 25 AM


- Only admin is able to add new members
- Only admin is able to change roles of any team member
- Only admin is able to delete any team member

- @slavingia please review the new changes.

@binary-koan so as of now we're banning our user by updating the ```app_metadata``` but now we have ```user_profiles``` table so,
1. we can add a column of isActive so when we want to remove any team member we can setIsActive to false .
2. We can remove the user and delete the whole user from user_profiles and also from supabase. Please share your opinion and any approach/idea.

Adarsh9977:
> We generally don't do booleans but non-null timestamps, so it'll be deleted_at

sound great, but IMO deleted_At should be nullable because, it will create problem while distinguish active and deleted users.

lib/data/user.ts
```diff
@@ -54,6 +55,40 @@ export const addUser = async (
...
+    .update(userProfiles)
+    .set({
+      deletedAt: new Date(),
+    })
+    .where(eq(userProfiles.id, userId));
+
+  if (error) {
+    throw error;
+  }
+};
```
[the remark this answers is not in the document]
Adarsh9977: Will catch that in new PR. Also we don't want to return anything so we use only await call as of now.

Adarsh9977:
@slavingia @binary-koan This PR is now ready for review. Done with all the comments from AI and team. updated ```user_metadata``` to ```user_profiles``` needed for this functionality. Will add a new PR for full ```user_metadata``` to ```user_profiles``` upgrade.

trpc/router/user.ts
```diff
@@ -12,13 +12,18 @@ import { env } from "@/lib/env";
...
-      where: eq(authUsers.email, input.email),
+    const user = await db.query.userProfiles.findFirst({
+      where: and(
+        eq(userProfiles.email, input.email),
+        isNull(userProfiles.deletedAt)
+      ),
     });
-    if (!user) {
+    
+    if (!user || user.deletedAt) {
```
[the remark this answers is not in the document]
Adarsh9977: This is added for double verification.

components/assignSelect.tsx
```diff
@@ -25,6 +25,8 @@ export const AssignSelect = ({ selectedUserId, onChange, aiOption, aiOptionSelec
   const [open, setOpen] = useState(false);
   const [searchTerm, setSearchTerm] = useState("");
   const [highlightedIndex, setHighlightedIndex] = useState(0);
+  const [localSelected, setLocalSelected] = useState<AssigneeOption | null | undefined>(undefined);
```
binary-koan, line 28: Are the changes here related? (if so how?)

db/schema/userProfiles.ts
```diff
@@ -9,8 +9,10 @@ export const userProfiles = pgTable("user_profiles", {
   id: uuid()
     .primaryKey()
     .references(() => authUsers.id, { onDelete: "cascade" }),
+  email: text(),
```
binary-koan, line 12: Why are we adding this here when we can get it from `authUsers`?

lib/data/user.ts
```diff
@@ -52,6 +53,43 @@ export const addUser = async (
...
+    error: getUserError,
+  } = await supabase.auth.admin.getUserById(userId);
+  if (getUserError) throw getUserError;
+
+  const existingAppMetadata = user?.app_metadata || {};
+
+  const { error } = await supabase.auth.admin.updateUserById(userId, {
+    app_metadata: {
+      ...existingAppMetadata,
+      banned: true,
```
binary-koan, line 78: Do we still need this?

trpc/router/user.ts
```diff
@@ -12,13 +12,18 @@ import { env } from "@/lib/env";
...
 import { publicProcedure } from "../trpc";
+import { userProfiles } from "@/db/schema";
 
 export const userRouter = {
   startSignIn: publicProcedure.input(z.object({ email: z.string() })).mutation(async ({ input }) => {
-    const user = await db.query.authUsers.findFirst({
-      where: eq(authUsers.email, input.email),
+    const user = await db.query.userProfiles.findFirst({
+      where: and(
+        eq(userProfiles.email, input.email),
```
binary-koan, line 21: Ah, is this why `email` was added? Would be better to join or use `with` rather than storing data in two places

app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx
```diff
@@ -30,14 +32,18 @@ const TeamSetting = ({ mailboxSlug }: TeamSettingProps) => {
     );
   });
 
+  const { data: conversationsList } = api.mailbox.conversations.list.useQuery({
```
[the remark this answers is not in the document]
Comment (Adarsh9977) [reply app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/teamSetting.tsx:35]: Yes, got your point.

trpc/router/user.ts
Adarsh9977: Can we update our trigger for that.So it will add email in ```user_profiles``` directly by trigger.

Adarsh9977: While working on replacing ```user_metadata``` PR I got this issue and we need email for signIn. So, best way to update our trigger.

app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/deleteMemberDialog.tsx
```diff
@@ -0,0 +1,155 @@
...
+        toast({
+          variant: "destructive",
+          title: "Please select a valid assignee",
+        });
+        setLoading(false);
+        return;
+      }
+
+      try {
+        const updatePromises = conversationIds.map((conversationId) => {
```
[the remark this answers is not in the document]
Comment (Adarsh9977) [reply app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/deleteMemberDialog.tsx:93]: @binary-koan  How about I can add a bulkAssign API at this moment because I have some followup/upcoming PR's after that. (If you allow)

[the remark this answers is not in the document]
Comment (binary-koan) [reply app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/deleteMemberDialog.tsx:93]: Fair enough, I think the best approach (if it works) is to add assignee parameters to the existing `bulkUpdate` endpoint and job

trpc/router/user.ts
binary-koan: Surely we can join when we need it? For example in signIn we could do

```ts
const user = await db
      .select({ email: authUsers.email, deletedAt: userProfiles.deletedAt })
      .from(authUsers)
      .innerJoin(userProfiles, eq(authUsers.id, userProfiles.id))
      .where(eq(authUsers.email, input.email))
      .then(takeUniqueOrThrow);
```

The point of linking users and userProfiles is that we don't need to duplicate data

app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/deleteMemberDialog.tsx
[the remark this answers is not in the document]
Comment (Adarsh9977) [reply app/(dashboard)/mailboxes/[mailbox_slug]/settings/team/deleteMemberDialog.tsx:93]: > Fair enough, I think the best approach (if it works) is to add assignee parameters to the existing `bulkUpdate` endpoint and job

ok let me try that

trpc/router/user.ts
Adarsh9977: > Surely we can join when we need it? For example in signIn we could do
> 
> ```ts
> const user = await db
>       .select({ email: authUsers.email, deletedAt: userProfiles.deletedAt })
>       .from(authUsers)
>       .innerJoin(userProfiles, eq(authUsers.id, userProfiles.id))
>       .where(eq(authUsers.email, input.email))
>       .then(takeUniqueOrThrow);
> ```
> 
> The point of linking users and userProfiles is that we don't need to duplicate data

Thanks, will use join whenever needed.

components/assignSelect.tsx
Adarsh9977: @binary-koan Yes, we need this for update the local state of assignSelect.

Adarsh9977:
@binary-koan This PR is finally tested ready for review and merge. IMO the reassignAll is the best approach to tackle the limit of conversationIds. Now we sending the count of all assigned tickets and reAssignAll api will update if count <= 25 else added a trigger Event which can be done by bulkUpdateConversations.

Issue #483, opened 2025-06-04
Common issues


Rachel-splatt:
## What

View groups of tickets about similar things

### Why

- Speed up support workflow by allowing users to answer or triage tickets about a single issue at once, rather than jumping between unrelated tickets. 
- Helps users spend less time searching for tickets they can help with

## Design (concept)

[Figma](https://www.figma.com/design/EuvcjpcXW72xG2pVtXku2G/Recent-activity--common-issues-and-nav-updates?node-id=1-3412&t=btrD7SmRt8MzvZxn-1)

Dev Mode: https://www.figma.com/design/EuvcjpcXW72xG2pVtXku2G/Recent-activity--common-issues-and-nav-updates?node-id=1-3412&m=dev&t=45RVMD6xQqrqW5fP-1

<img width="1495" alt="Image" src="https://github.com/user-attachments/assets/a8e9f660-1b13-455c-a31c-85ae551c6c01" />

chuporustum:
hey @slavingia 
Can you make the figma for this public please? This sounds super fun and I think I can ship this quickly.

chuporustum:
opened a PR for this^

slavingia:
If you request Figma Dev mode access, that may help nail all of the values.

chuporustum:
> If you request Figma Dev mode access, that may help nail all of the values.

I did requested that day itself (as notified [here](https://github.com/antiwork/helper/issues/483#issuecomment-3039277813)) but didn't get yet. Could you please help with this?

chuporustum:
> Added a Dev Mode link to the description.

Still not able to access the same

<img width="718" height="378" alt="Image" src="https://github.com/user-attachments/assets/8c331114-0e6c-4d67-9332-6b46a1cc6b15" />

slavingia:
Email? Will manually add.

chuporustum:
> Email? Will manually add.

Please use chuporustum@gmail.com

slavingia:
invited, sorry for the delay

PR #578, opened 2025-06-24
feat: introduce grouping by common issue topics

debadree25:
Tries to address: https://github.com/antiwork/helper/issues/483

Hello!

In this PR have started working on implementing the common issue grouping feature, the present version includes the backend implementation heres what I have done:

- Introduce a issueTopic table linking it to conversations
- Introduced a cron which runs for all non-classified conversations summarises them and populates and connects them to the issueTopics and also stores the embedding for the same 

Would like you thoughts on this and whether to pursue this line of implementation?

Thank you!

## Summary by CodeRabbit

- **New Features**
	- Conversations are now automatically grouped by concise issue topics, enabling easier organization and retrieval.
	- A background process classifies ungrouped conversations by topic every 5 minutes.
- **Enhancements**
	- Improved search efficiency for similar conversation topics using advanced vector indexing.
	- Expanded tracking of AI usage events to include issue topic generation.


binary-koan:
I think it's best if we scope this more internally, sorry. We haven't had good results with AI categorisation in the past so a better approach may be to allow users to define issues and have AI just put tickets into those buckets - but still thinking on that.

Issue #593, opened 2025-06-25
Filtered empty state message should show independently from onboarding/setup dialog


MuskanPaliwal:
## Problem
When a filter or search is applied in the mailbox conversation list, the empty state message for no results should always be shown, even if the mailbox is not fully set up (onboarding). Previously, the onboarding dialog would always take precedence, hiding the filtered empty state message.

## Solution- Show the filtered empty state message ("No conversations found for this filter") whenever a filter or search is applied and there are no results, regardless of onboarding state.
- Show the onboarding/setup dialog only when there are no filters/search and the mailbox is not set up.

## Acceptance Criteria
- [x] When a filter/search is applied and there are no results, the user sees the filtered empty state message.
- [x] When there are no filters/search and the mailbox is not set up, the user sees the onboarding/setup dialog.
- [x] When there are no open tickets and no filters, the user sees the "No open tickets" message.
- [x] All debug logs are removed.

I am still confused if the logic is correct but this should be the expected behavior. Not sure where to put the setup dialog after this. Would love to discuss. @slavingia

PR #594, opened 2025-06-25
Show Filtered Empty State Message Independently from Onboarding/Setup Dialog

MuskanPaliwal:
### Summary

This PR updates the mailbox conversation list empty state logic to improve user experience when filters or search are applied:

- **Filtered empty state message** ("No conversations found for this filter") is now shown whenever a filter or search is active and there are no results, regardless of the mailbox onboarding/setup state.
- **Onboarding/setup dialog** is only shown when there are no filters/search and the mailbox is not fully set up.
- **No open tickets** and **default empty state** messages remain unchanged for their respective scenarios.
- All debug logs have been removed.

### Motivation

Previously, the onboarding dialog would always take precedence, even when a user applied a filter or search. This could be confusing, as users would not see feedback that their filter/search returned no results. Now, the UI always provides clear feedback for filtered empty states, improving clarity and usability.

### Acceptance Criteria

- [x] When a filter/search is applied and there are no results, the user sees the filtered empty state message.
- [x] When there are no filters/search and the mailbox is not set up, the user sees the onboarding/setup dialog.
- [x] When there are no open tickets and no filters, the user sees the "No open tickets" message.
- [x] All debug logs are removed.

### Related Issue

Closes [#593](https://github.com/antiwork/helper/issues/593)

## Summary by CodeRabbit

- **New Features**
  - Improved empty state messaging in conversations list to clearly indicate when no conversations match the current filters or search criteria.

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
@@ -239,7 +239,7 @@ export const List = () => {
           <LoadingSpinner size="lg" />
         </div>
       ) : conversations.length === 0 ? (
-        <NoConversations filtered={Object.values(filterValues).some(Boolean) || !!input.search} />
+        <NoConversations filtered={activeFilterCount > 0 || !!input.search} />
       ) : (
         <div ref={resultsContainerRef} className="flex-1 overflow-y-auto">
           {conversations.map((conversation) => (
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
@@ -23,6 +23,7 @@ import NewConversationModalContent from "./newConversationModal";
 type ListItem = ConversationItem & { isNew?: boolean };
 
 export const List = () => {
+  console.log("[ConversationList] Rendering List component");
   const [conversationSlug] = useQueryState("id");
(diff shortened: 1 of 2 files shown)


app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
```diff
@@ -239,7 +239,7 @@ export const List = () => {
           <LoadingSpinner size="lg" />
         </div>
       ) : conversations.length === 0 ? (
-        <NoConversations />
+        <NoConversations filtered={Object.values(filterValues).some(Boolean) || !!input.search} />
```
Comment (binary-koan) [inline app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx:242]:
```suggestion
        <NoConversations filtered={activeFilterCount > 0 || !!input.search} />
```

binary-koan (commented):
One suggestion - could you also add a screenshot of the new state for reference?

This behaviour looks good to me for now. We should probably move the onboarding state to the sidebar now that there's tons of space there actually, but not urgent and should be done separately

MuskanPaliwal:
> One suggestion - could you also add a screenshot of the new state for reference?
> 
> This behaviour looks good to me for now. We should probably move the onboarding state to the sidebar now that there's tons of space there actually, but not urgent and should be done separately
Right makes sense. If this is not urgent then I can pick up this later or I maybe I will create a issue for the same.

MuskanPaliwal:
Here is a recording for the current state @binary-koan 

https://github.com/user-attachments/assets/20acb586-57b8-4373-8b1b-e7ca9eb4a87f

PR #600, opened 2025-06-25
chore: Removed unused code

aparna2071:
Removed a bunch of unused code, there should be no functional changes.

## Summary by CodeRabbit

- **Removed Features**
  - Several dropdown menu components and related UI elements have been removed.
  - The ability to close sheets using a specific component has been removed.
  - The global Escape key handler hook is no longer available.
  - Telemetry and error tracking setup, including request error capturing, has been removed.
  - The GitHub redirect URI constant is no longer present.
- **Chores**
  - Unused type declarations have been removed from the global type definitions.


components/ui/dropdown-menu.tsx
```diff
@@ -8,45 +8,9 @@ import { cn } from "@/lib/utils";
...
-  <DropdownMenuPrimitive.SubContent
-    ref={ref}
-    className={cn(
-      "z-50 min-w-[8rem] overflow-hidden rounded-md border bg-popover p-1 text-popover-foreground shadow-lg data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
-      className,
-    )}
-    {...props}
-  />
-));
-DropdownMenuSubContent.displayName = DropdownMenuPrimitive.SubContent.displayName;
```
slavingia, line 49: I don't think we should change this as it's a Shadcn component

slavingia (changes requested):

binary-koan (commented):
Added a few comments, could you check?

instrumentation.ts
```diff
@@ -1,19 +0,0 @@
-import * as Sentry from "@sentry/nextjs";
```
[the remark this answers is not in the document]
aparna2071: Okay, will revert back the whole instrumentation.ts file.

components/ui/dropdown-menu.tsx
binary-koan: Please revert all changes in components/ui, not just this file. Those are all shadcn components where we should stay as close as possible to the original template

aparna2071: Ok thanks @binary-koan, will do

binary-koan:
This ended up overlapping with #2150 - I think that one removed basically all the same things so will close this. Sorry about that

PR #611, opened 2025-06-27
Cleanup: Remove unused exports, types, dependencies, and add knip.json

apdev2:
- Removes unused exports to reduce bundle size and simplify the codebase
- Deletes unused files:
   - `lib/github/constants.ts`
   - `trpc/serverActions.ts`
   - `components/useOnGlobalEscPress.ts`
- Removes unused dependencies:
   - `@radix-ui/react-hover-card`
   - `react-sortablejs`
- Adds `.tspruneignore` to exclude known safe files and directories from `ts-prune` checks (`**/middleware.ts`, `**/instrumentation.ts`, `**/app/global-error.tsx`, etc.).

## Summary by CodeRabbit

* **Chores**
  * Removed several unused types, constants, and dependencies to streamline the codebase.
  * Deleted configuration and utility files that are no longer needed.
  * Updated visibility of various internal functions and constants to reduce unnecessary exports.
  * Added a configuration to ignore specific UI component files in tooling.
* **Documentation**
  * No user-facing documentation changes included.


.tspruneignore
```diff
@@ -0,0 +1,8 @@
+public
```
binary-koan, line 1: ts-prune is deprecated, we should use knip (and add an equivalent knip.json or similar)

components/ui/avatar.tsx
```diff
@@ -54,5 +54,3 @@ export const Avatar = ({ src, fallback, size = "md" }: CustomAvatarProps) => {
     </BaseAvatar>
```
binary-koan, line 54: Please revert changes to components/ui and (if possible) configure knip to ignore unused exports in these files. They're shadcn so we want to stay as close to the template as possible

binary-koan (commented):
Thanks! Added a couple of comments

components/ui/avatar.tsx
apdev2: Reverted the change and configured to ignore these in knip.json.

Would you prefer having a git pre-commit hook that logs a warning (and continues?), whenever someone makes changes to `components/ui`?

apdev2:
Hello, Just wanted to confirm - does @slavingia's [tweet](https://x.com/shl/status/1935719392630698461) apply to this PR?

apdev2:
Hi @slavingia, I've reached out over the above email. Can you please check?

apdev2:
Hi @slavingia, I've reached out over the above email. Can you please check?

PR #617, opened 2025-06-29
refactor: follow naming convention for naming files

Udit-takkar:
No change in functionality. 
Just changed the name of files to follow camelCase naming convention as followed in the codebase

## Summary by CodeRabbit

* **Chores**
  * Updated internal references for several utilities and hooks to reflect new module paths. No user-facing changes or functional differences are introduced.


app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/newConversationModal.tsx
```diff
@@ -7,8 +7,8 @@ import {
 import { EmailSignature } from "@/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/emailSignature";
 import { DraftedEmail } from "@/app/types/global";
 import { FileUploadProvider, useFileUpload } from "@/components/fileUploadContext";
-import { toast } from "@/components/hooks/use-toast";
 import { useSpeechRecognition } from "@/components/hooks/useSpeechRecognition";
+import { toast } from "@/components/hooks/useToast";
```
Comment (slavingia) [inline app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/newConversationModal.tsx:11]: Why reorder?

slavingia (changes requested):

Udit-takkar:
>Revert useToast to use-toast since it's part of shadcn (speaking of which, we should switch to their new sonner toast as this one is deprecated - no need to do here though)

done

>Move the other hooks out of the hooks folder - we typically don't use those and put hooks alongside components, so makes more sense to reserve that folder for shadcn IMO

okay moved the hooks that weren't used by shadcn components

binary-koan (commented):
I don't think `useNow` and `useSavingIndicator` should be in `ui`?

PR #623, opened 2025-06-30
Fix: Improve Clear Formatting Button UX in Rich Text Editor

chuporustum:
## 📋 **Summary**
Transforms the confusing "Tx" clear formatting button into a smart, contextual tool that only appears when genuinely useful, eliminating user confusion and improving efficiency.

## 🚨 **Problem**
The current clear formatting button had critical UX issues:
- **Always visible** even when not needed (redundant for single formatting)
- **Poor visual communication** - "Tx" icon didn't convey purpose  
- **Appeared non-responsive** when clicked on unformatted text
- **Misleading hints** confused users about actual functionality

## ✅ **Solution**
- **Smart visibility**: Button only appears when multiple formats are applied to selected text
- **Professional icon**: Replaced confusing "Tx" with proper RemoveFormatting icon
- **Real-time updates**: Dynamically shows/hides based on current selection
- **Cleaner interface**: Removed misleading hint bubbles

### Before (Issues)

https://github.com/user-attachments/assets/c601f0db-1d99-4fb1-aeba-b4205f1bb20d

### After (Improved)  

https://github.com/user-attachments/assets/801bad84-da64-4042-98e5-d9f92aa616e6

## 📊 **Impact**
- **Eliminated** user confusion about button purpose

## Summary by CodeRabbit

* **New Features**
  * The "Clear formatting" button now appears only when multiple text formats are applied to a selection.
  * Added a tooltip to the "Clear formatting" button that displays "Clear all formatting" on hover.

* **Bug Fixes**
  * Improved toolbar responsiveness to selection changes for more accurate button display.

* **Removed**
  * The in-editor "Hint: Paste URL to create link" bubble menu has been removed.


binary-koan:
I like the idea of adding tooltips (we should do it for all the toolbar buttons actually) but I'm not so keen on conditionally showing the button - I think the improvement isn't worth the complexity this adds to the editor. If we had a contextual bubble menu I could get behind hiding and showing buttons, but since the toolbar is static I think it's best we keep it consistent.

PR #632, opened 2025-06-30
feat (conversations): recent searches popup

omeraydindev:
This PR introduces a 'recent searches' feature in the conversations page. The user can also delete their recent searches one by one.

The PR also adds a new X button on the search bar to clear the input which is convenient.

Do let me know if there is anything else I need to do regarding the schema change/DB migration, or if I need to add more E2E tests! 

https://github.com/user-attachments/assets/d811601d-303f-4746-87de-abe97846193a

## Summary by CodeRabbit

* **New Features**
  * Added a recent searches dropdown to the conversation search bar, allowing users to quickly select or delete previous search terms.
  * Recent searches are saved per mailbox and can be managed directly from the search input.

* **Tests**
  * Introduced end-to-end tests to verify recent search selection and deletion functionality in the conversation search bar.


binary-koan (commented):
I like the idea, but I'm not sure it's worth a new db table - I think just putting it in localStorage would solve the majority of use cases since most people will be using Helper on a specific work computer most of the time. wdyt about just doing this client side for now?

omeraydindev:
@binary-koan The client side part should be all good, do let me know if there is any other action you need me to take!

Just as a side note, if this gets merged, a squash and merge might be more appropriate if we don't want to have the -now removed- backend part of this PR to show up in the commit history.

binary-koan (commented):
I tried this out and I think it would work better if we

* Make it trigger when clicking the "x" button to clear the search field
* Stop the search bar autofocusing on page load so that you need to explicitly focus it and trigger the popup

Mind updating that?

omeraydindev:
@binary-koan Done, good calls! Those really improved the UX for this feature

slavingia:
Don't think this is valuable personally

PR #637, opened 2025-07-01
Fix TipTap link modal stacking and positioning

### LINKED ISSUE #602: Link Modal Popover: Stacking and Positioning Issue
**Problem:**  
The link modal in the TipTap editor can be hidden behind dialogs/overlays and is not always centered above the link icon.

**Current Behaviour:**  
- Modal may be hidden or misaligned
- Adding screenshot for the reference:
Image

**Expected:**  
- [ ] Modal always appears above overlays
- [ ] Modal is centered above the link icon
Image

**Impact:**  
Users can’t reliably access the link modal, affecting editor usability.

Anshgrover23:
Fixes #602 

### Reference Screenshots: 
Before: 
image

After:
photo_2025-07-01_14-56-56


binary-koan (commented):
Sorry for the delay! Could you explain why simpler z-index adjustments don't work for this - why we need to use a portal?

Anshgrover23:
@binary-koan hey, thanks for the review, here is the explanation:
Using a portal is necessary because z-index adjustments alone cannot guarantee the modal will appear above all overlays. This is due to stacking context limitations in CSS—if the modal is rendered inside a parent with its own stacking context, it can be visually trapped beneath other elements, regardless of z-index. By rendering the modal in a portal at the document body level, we ensure it is not constrained by parent stacking contexts and will always appear above global overlays and dialogs.

Anshgrover23:
@slavingia can i get review here as well?

slavingia:
Don't need one from me

PR #644, opened 2025-07-01
perf: optimize list components with React.memo

aKrhn:
This performance optimization adds React.memo to three heavy list components that were causing unnecessary re-renders in large datasets:

Components optimized:
- ConversationListItem: Now only re-renders when conversation data changes
- TeamMemberRow: Now only re-renders when member data changes
- MessageItem: Now only re-renders when message content changes

Changes:
- Added React.memo wrappers with custom comparison functions
- Optimized prop comparison to avoid deep object comparisons where possible
- Added displayName for better debugging experience

Performance improvements:
- Reduced re-renders in conversation lists with 100+ items
- Improved team settings page performance with large teams
- Smoother message thread scrolling during real-time updates
- Lower CPU usage during live conversation updates

Impact:
- Users will notice smoother scrolling and interactions
- Reduced battery drain on mobile devices
- Better performance with large datasets
- Improved user experience in busy workspaces

## Summary by CodeRabbit

* **Refactor**
  * Improved performance of message, conversation, and team member list items by optimizing their rendering with advanced memoization techniques.
  * Enhanced component structure for better maintainability and debugging without affecting existing functionality or user interface.

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/messageItem.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/messageItem.tsx
@@ -1,5 +1,5 @@
 import cx from "classnames";
-import { useMemo, useState, type JSX } from "react";
+import { memo, useMemo, useState, type JSX } from "react";
 import type { AttachedFile, Conversation, Message as MessageType, Note as NoteType } from "@/app/types/global";
 import HumanizedTime from "@/components/humanizedTime";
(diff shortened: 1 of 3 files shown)


binary-koan:
I appreciate the idea, but I don't think this approach is worth it unless we can point to extreme real-world performance issues - especially with React Compiler on the horizon. Having detailed checks like this separate from the component logic makes it brittle and easy for humans and AI to introduce bugs (and less intuitive to debug)

PR #663, opened 2025-07-01
fix: improve conversation UI positioning and resize behavior

hbce:
## Before Video

https://github.com/user-attachments/assets/cb57d759-febc-4427-9bb0-dd6a0b0aee5a

## After Video

https://github.com/user-attachments/assets/ae82d9ee-c981-49b5-afb5-ba8436cfbe21

## ScrollToTop Button Improvements
- Change from absolute to sticky positioning for better visibility
- Move ScrollToTop button inside MessageThreadPanel for proper context
- Ensure button stays visible and correctly positioned during resizes

## Enhanced Panel Resize Logic
- Add smart scroll behavior that maintains bottom position during resize
- Auto-scroll to bottom when user was already at bottom before resize
- Use 50px threshold to detect if user is at conversation bottom

## Better Panel Size Constraints
- Update maxSize from 80 to 62.5 for better balance between panels
- Add minSize constraint of 37.5 for message actions panel
- Add maxSize constraint of 80 for message actions panel
- Prevent extreme resize scenarios that make interface unusable

## Message Actions Panel Enhancement
- Add overflow-y-auto for independent scrolling when content is tall
- Better handling of long compose/reply content

These changes fix several UX issues:
- ScrollToTop button no longer disappears during resize
- Conversation stays scrolled to bottom when user resizes panels
- Better balance between message thread and compose areas
- Independent scrolling for compose area when needed

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/conversation/conversation.tsx
@@ -131,7 +131,7 @@ const ScrollToTopButton = ({
       <TooltipTrigger asChild>
         <a
           className={cn(
-            "absolute bottom-4 left-4 transition-all duration-200 h-8 w-8 p-0 rounded-full",
+            "sticky bottom-4 left-4 z-10 transition-all duration-200 h-8 w-8 p-0 rounded-full",
             "flex items-center justify-center",
             "bg-background border border-border shadow-xs",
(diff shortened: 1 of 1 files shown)


hbce:
I noticed that today is the last day, could you please let me know the final status regarding @slavingia post? https://x.com/shl/status/1935719392630698461

thank you for all.

hbce:
sorry but it is a bug and i fixed it, considering the your post in x, it deserves payment

PR #668, opened 2025-07-02
feat: improve conversation list selection styling and sort dropdown

hbce:
Before Video:

After Video:

https://github.com/user-attachments/assets/35386576-1e7e-4290-8e0e-a79576d6ced7

https://github.com/user-attachments/assets/98cbb670-5b53-4438-a706-1b5abf1f36a7

- Add visual indication for selected conversations with secondary background
- Fix sort dropdown to always show selected option properly
- Improve sort option selection logic with useEffect
- Simplify sort button display for better consistency
- Import additional sort icons for future use


binary-koan:
This looks overcomplicated for what it is tbh - I think it would simplify things a lot if we don't render the sort dropdown until `defaultSort` has loaded, could you try that?

Also the design improvements in the PR description seem to be gone - was that intentional?

PR #678, opened 2025-07-02
feat: add closed by filter (ai or human), trigger from the statusByTypeChart

omeraydindev:
This PR adds a nice "Closed by" filter that is visible when the status filter is set to "Closed". It lets the user filter whether the ticket was closed manually or by AI.

It also integrates the Ticket Status chart in the dashboard so that whenever the user clicks on a slice, they're redirected to the conversations page with the filter already applied.

Demo:

https://github.com/user-attachments/assets/f4f6290d-d639-422a-bbd8-fc8bf811b421


slavingia:
Did this come from a user request? Im not sure if this is useful personally but open to consider

slavingia:
IMO not worth the LOC

omeraydindev:
It's your call, but just to share my perspective - I genuinely think the Ticket Status chart loses a lot of its value if it's not even clickable and gives the customer zero details. This PR was mainly about making it more useful.

PR #682, opened 2025-07-03
fix: Resolve memory leak in useRealtimeEventOnce hook

VasuDevrani:
### PROBLEM
The handledOneTimeMessageIds Set is global. If the component unmounts and remounts, previous IDs remain leading to excessive memory usage.

### SOLUTION
- Memory leak prevention: Set is now capped at 500 entries and old entries expire after 5 minutes.
- TTL cleanup: This avoids unchecked growth in long sessions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance and reliability of real-time event handling by automatically removing old message IDs, preventing potential memory issues during extended use.

### DIFF
--- a/lib/realtime/hooks.ts
+++ b/lib/realtime/hooks.ts
@@ -43,15 +43,29 @@ export const useRealtimeEvent = <Data = any>(
   }, [channel, event]);
 };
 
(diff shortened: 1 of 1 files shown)


VasuDevrani:
@binary-koan any suggestions here?

Issue #489, opened 2025-06-05
Remove mailbox switching


binary-koan:
- [x] Move Flexile and Helper mailboxes to separate instances
- [x] Remove the mailbox switcher dropdown and the mailboxes.list procedure
- [x] Replace `getMailboxById` and `getMailboxBySlug` with a `getMailbox` function that just returns the first mailbox
- [x] One by one, rename `mailboxId` db fields to `unused_mailboxId` and remove their usage
- [x] Remove `useConversationListContextSafe` and associated bloat
- [ ] Remove Flexile and Helper mailboxes from seed data so we just create Gumroad
- [x] Remove the /mailboxes/:slug prefix from all routes (@binary-koan)

### Why

Now that we're single tenant we can just spin up a new instance per mailbox. We don't have proper mailbox CRUD anyway so this makes the code less confusing.

slavingia:
If you're open to it, setting up Flexile (including user metadata and tools use) may make for a good/great tutorial video on how to get helper setup

binary-koan:
Open to contributions on this!

kcansari:
I’d like to contribute to the second sub-issue titled `Remove the mailbox switcher dropdown and the mailboxes.list procedure` @binary-koan

Anshgrover23:
@slavingia Can I take this up since no one is working on it?

Anshgrover23:
@binary-koan `Remove useConversationListContextSafe and associated bloat` for this part can u simplify wdym by bloat?

areu trying to say remove the usage of `mailboxSlug` from the UI right?

binary-koan:
Yeah if we remove `mailboxSlug` from the UI everywhere it'll solve that one - I think it was more just a reminder to remove that function entirely once we don't need it any more, sorry for the weird wording

Anshgrover23:
@binary-koan just a little thing i wanted to know, do i have to edit these hrefs like removing mailboxes completely or not, also do i need to do any change in the schema file of mailboxes regarding this? can u clarify?

<img width="798" height="152" alt="Image" src="https://github.com/user-attachments/assets/aa3c7c08-5c10-4b86-b5d4-5056419e2337" />

Anshgrover23:
@slavingia got it on removing the mailboxSlug from the routes. Just to confirm , do you also want me to remove the mailboxes table or related schema references, or should I leave the schema untouched for now?

binary-koan:
[...]

PR #612, opened 2025-06-27
fix(search-conversations): incorrect rendering of highlighted text in body

kcansari:
Problem Solved

  I fixed an issue with conversation search body highlighting not 
  displaying correctly. The problem was that when users searched for
  terms that appeared deep within message text, the default text
  truncation would cut off the search results, making the highlighted
  matches invisible.

Helper

  Solution Summary

  I implemented a smart search snippet system that:

  1. Created a new utility (lib/search/searchSnippet.ts) that generates
  context-aware text snippets around search matches
  2. Enhanced the conversation list item to use snippets when search
  terms appear beyond the normal truncation point
  3. Added comprehensive E2E tests to ensure the feature works correctly
   across different scenarios

Mine

  Key Features Implemented

  Smart Snippet Logic

  - Early match detection: If search terms appear early enough that
  normal truncation would show them, keeps original behavior
  - Deep match handling: For matches that appear later in the text,
  creates a snippet centered around the match with "..." indicators
  - Context preservation: Shows surrounding words for better readability
  - Word boundary respect: Avoids cutting words in half when possible

  UI Improvements

  - Dynamic styling: Removes truncate class and adds leading-relaxed for
   snippets
  - Word breaking: Adds word-break: break-word for better snippet
  display
  - Maintains highlighting: Search terms remain highlighted within
  snippets

  Test Coverage

  Added 5 comprehensive E2E tests covering:
  - Truncated text for non-search results
  - Early match behavior preservation
  - Context snippets for deep matches
  - Search term highlighting in snippets
  - Graceful handling of no results

  This solution significantly improves the search experience by ensuring
   users can always see their search results, regardless of where they
  appear in the message text.

## Summary by CodeRabbit

* **New Features**
  * Improved display of conversation search results by showing context-rich snippets with highlighted search terms.
  * Added ellipsis to indicate truncated snippets when search matches are found deeper in the message.

* **Bug Fixes**
  * Ensured keyword highlighting appears only when search terms are present in the matched message body.

* **Tests**
  * Introduced new end-to-end tests to verify search snippet display, keyword highlighting, and UI behavior for various search scenarios.

PR #685, opened 2025-07-03
feat: remove mailbox switcher dropdown and mailboxes.list procedure | Remove mailbox switching #489

kcansari:
I implemented the subtask based on my understanding of the feature. If there’s anything you’d like me to adjust or improve, I’d be happy to update it.

#489 for second sub-issue -> Remove the mailbox switcher dropdown and the mailboxes.list procedure

before:
before-dropdown

after:
after-dropdown

## Summary by CodeRabbit

* **New Features**
  * Simplified mailbox sidebar to display only the current mailbox name and avatar without a dropdown or switching option.

* **Refactor**
  * Improved mailbox validation and navigation by directly fetching mailboxes by slug instead of retrieving the full mailbox list.
  * Updated cache invalidation to target only the current mailbox after updates.

* **Bug Fixes**
  * Enhanced error handling and redirection when accessing mailboxes that do not exist.

* **Tests**
  * Removed tests related to listing all mailboxes.

* **Chores**
  * Removed unused code, imports, and the mailbox list query from the backend.

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/appSidebar.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/appSidebar.tsx
@@ -91,7 +91,7 @@ export function AppSidebar({ mailboxSlug }: { mailboxSlug: string }) {
           </SidebarMenu>
         ) : (
           <div className="flex items-center gap-2 w-full h-10 px-2 rounded-lg">
-            <Avatar src={undefined} fallback={mailbox?.name || ""} size="sm" />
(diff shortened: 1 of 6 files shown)


slavingia:
This does the front-end but the value of the PR would be massively simplifying the whole codebase to be single tenant

kcansari:
> This does the front-end but the value of the PR would be massively simplifying the whole codebase to be single tenant

> I see this is the recommended approach in the PR atomically so please proceed!

Okay, I’ve just solved the issue you mentioned in #612, and I’ll continue working on this task.

PR #687, opened 2025-07-03
fix: removed unused shadcn components and add confirmation dialog in saved replies

Adarsh9977:
### Removed unused shadcn components and add inbuilt confirmation dialog instead of alert 

https://github.com/user-attachments/assets/419a66a9-0f41-4c27-b22e-f06d94671046

## Summary by CodeRabbit

* **Refactor**
  * Simplified the confirmation dialog for deleting saved replies, replacing the previous multi-component dialog with a more streamlined version.
* **Chores**
  * Removed unused alert dialog and scroll area UI components from the codebase.


slavingia:
Am I wrong for wanting to go the other way? Get rid of our component and use shadcn?

@binary-koan

Adarsh9977:
@slavingia I am with you but, our inbuilt component ensures that all our confirmation dialogs follow the same pattern, design and styling because we use confirmation dialog very often everywhere. It maintain a consistency in our app.

slavingia:
Wouldn't using shadcn do the same (since AI can see other uses) and allow for more customization when we need it?

slavingia (approved):

Adarsh9977:
> Wouldn't using shadcn do the same (since AI can see other uses) and allow for more customization when we need it?

Yeah you're right. Raw shadcn is very much customizable by AI instead of any inbuilt wrapper. Got your point now.

PR #697, opened 2025-07-04
fix: date-time formatting dashboard

VasuDevrani:
### NOW 

image

### BEFORE 

image


VasuDevrani:
> This bypasses the `formatter.days()` and `formatter.hours()` functions used in other cases, breaking the formatter abstraction.

@binary-koan i have given a simple 1 line fix, should i address the cursor ai comment above

binary-koan (approved):
I feel like this function could use a rework, but this is a good fix for now!

PR #707, opened 2025-07-05
fix (conversations): select all not bulk updating all items

omeraydindev:
This PR fixes an issue where clicking "Select all" and performing a bulk update only sends the currently loaded items to the backend which causes incomplete updates.

The fix was inspired by @binary-koan's [comment](https://github.com/antiwork/helper/pull/627#discussion_r2179514057) in a separate discussion, noting that the endpoint is expected to support bulk actions using both filters and ID lists when "Select all" is used.

Before:

https://github.com/user-attachments/assets/98b7c7d5-302e-4082-9026-494990fc9a32

After:

https://github.com/user-attachments/assets/b064b924-6b15-49ef-96d8-d1ecd8553414

### DIFF
--- a/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
+++ b/app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
@@ -82,8 +82,7 @@ export const List = () => {
   const handleBulkUpdate = (status: "open" | "closed" | "spam") => {
     setIsBulkUpdating(true);
     try {
-      const conversationFilter = allConversationsSelected ? conversations.map((c) => c.id) : selectedConversations;
-      const selectedCount = allConversationsSelected ? conversations.length : selectedConversations.length;
+      const conversationFilter = allConversationsSelected ? input : selectedConversations;
 
       bulkUpdate(
         {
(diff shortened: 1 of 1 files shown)


binary-koan (commented):
We have a special case in the router to update conversations immediately if you pass <25 IDs (meant to be <= I think, i.e. one page of results). Would it work to add a case on the frontend to send the IDs if there are 25 or less total results so it triggers that logic?

https://github.com/antiwork/helper/blob/f2a5834ec1f849e77593c03a565edd4f80dfd7c1/trpc/router/mailbox/conversations/index.ts#L178

omeraydindev:
@binary-koan Yep that makes sense from a performance endpoint, just pushed a commit to do exactly that

app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx
```diff
@@ -82,8 +82,15 @@ export const List = () => {
   const handleBulkUpdate = (status: "open" | "closed" | "spam") => {
     setIsBulkUpdating(true);
     try {
-      const conversationFilter = allConversationsSelected ? conversations.map((c) => c.id) : selectedConversations;
-      const selectedCount = allConversationsSelected ? conversations.length : selectedConversations.length;
+      // If all conversations are selected:
+      //   - If there are 25 or fewer conversations and there is no next page, use the selected conversation IDs
+      //   - Otherwise, use the search filters for the backend to match the conversations
+      // If not, use the selected conversation IDs
```
Comment (binary-koan) [inline app/(dashboard)/mailboxes/[mailbox_slug]/[category]/list/conversationList.tsx:88]:
```suggestion
```

I don't think we need this comment, the code is readable enough

binary-koan (approved):
Looks great! I don't think we need the comment, I'll remove it before merging if you're OK with that

PR #722, opened 2025-07-08
feat: update test environment setup and scripts for e2e testing

dixitt5:
**Summary:**

- Create a bash script to run e2e tests in separate test 
- Update commands in package.json to accommodate the testing setup changes.

## Summary by CodeRabbit

* **New Features**
  * Added sample environment files for local development and testing to simplify setup.
  * Introduced a script to automate end-to-end test setup and execution.

* **Chores**
  * Updated environment variable management for local development and testing.
  * Standardized and simplified test scripts for consistency.
  * Made configuration files use environment variables for improved flexibility.
  * Improved Playwright and Supabase integration for local testing environments.

* **Bug Fixes**
  * Ensured URLs and ports in tests and configurations dynamically use environment variables, reducing hardcoded values and potential conflicts.


dixitt5:
@binary-koan had some questions regarding this setup:
1. do we want to spin us nginx container for proxy too when running e2e tests or just testing on localhost is fine?
2. running the bash script will remove the containers for supabase dev instance, didn't find any other solution for doing it separately. one way was to create a separate supabase project but that would require us mapping different (available) ports for every services.

binary-koan:
> do we want to spin us nginx container for proxy too when running e2e tests or just testing on localhost is fine?

Let's use localhost for simplicity and reconsider if we run into problems with it

> running the bash script will remove the containers for supabase dev instance, didn't find any other solution for doing it separately. one way was to create a separate supabase project but that would require us mapping different (available) ports for every services.

I think it's better to use a separate project with different port mappings (hopefully Supabase has a way to configure them without too much effort?) since we want to make it easy to run tests locally alongside the dev server

binary-koan:
I haven't run into any Supabase vector issues I'm afraid (or heard of people running into them)

dixitt5:
@binary-koan can you review the changes and let me know if I am going with the right direction?
by the current changes, both dev and testing env can exist together because of both project ids and ports being different therefore no chances of conflicts. 

I am facing one bug after these changes, somehow able to run the dev setup but getting a cors error when otp is getting verified on user login.

.env.local.sample
```diff
@@ -8,6 +8,18 @@ GOOGLE_CLIENT_SECRET="set_me"
 GOOGLE_PUBSUB_CLAIM_EMAIL="set_me@example.com"
 GOOGLE_PUBSUB_TOPIC_NAME="set_me"
 
+# Supabase configuration for local development
```
binary-koan, line 11: Can we put this in a `.env.development` and commit it so people don't have to look through these when creating their .env.local?

dixitt5:
> Do we not reset the database anywhere? Ideally, we should find a fast way to reset it that we can do between every test (pnpm db:reset is currently super slow).

For now, I’ve added supabase db reset before running migrations in the test.sh file. I also experimented with other approaches, like truncating all tables and reseeding the database before each test. However, I ran into issues—resetting the DB properly seems to require more than just truncating and seeding, as there are other stateful aspects that need to be handled for a truly clean slate.

If having a fast, reliable DB reset between tests is important for this PR, I’m happy to hand this off to someone else who wants to take a shot at it, since I hit a bit of a roadblock here.

@binary-koan

MuskanPaliwal:
> > Do we not reset the database anywhere? Ideally, we should find a fast way to reset it that we can do between every test (pnpm db:reset is currently super slow).
> 
> For now, I’ve added `supabase db reset` before running migrations in the `test.sh` file. I also experimented with other approaches, like truncating all tables and reseeding the database before each test. However, I ran into issues—resetting the DB properly seems to require more than just truncating and seeding, as there are other stateful aspects that need to be handled for a truly clean slate.
> 
> If having a fast, reliable DB reset between tests is important for this PR, I’m happy to hand this off to someone else who wants to take a shot at it, since I hit a bit of a roadblock here.
> 
> @binary-koan

Hi, I would love to take a shot at this.

.env.test
```diff
@@ -0,0 +1,13 @@
...
+SUPABASE_PROJECT_ID=helper-test
+SUPABASE_API_PORT=5544
+SUPABASE_DB_PORT=5545
+SUPABASE_DB_SHADOW_PORT=5546
+SUPABASE_STUDIO_PORT=5547
+SUPABASE_INBUCKET_PORT=5548
+SUPABASE_ANALYTICS_PORT=5549
+SUPABASE_DB_POOLER_PORT=5550
+
+# Your other environment variables go here as needed. refer to the .env.test.local.sample for more details.
```
binary-koan, line 13:
```suggestion
# Other environment variables and overrides can be added in .env.test.local. Refer to .env.test.local.sample for more details.
```

.env.test.local.sample
```diff
@@ -0,0 +1,24 @@
...
+# The project ID should be unique to avoid conflicts with other instances.
+SUPABASE_PROJECT_ID=helper-test
+SUPABASE_API_PORT=5544
+SUPABASE_DB_PORT=5545
+SUPABASE_DB_SHADOW_PORT=5546
+SUPABASE_STUDIO_PORT=5547
+SUPABASE_INBUCKET_PORT=5548
+SUPABASE_ANALYTICS_PORT=5549
+SUPABASE_DB_POOLER_PORT=5550
+
```
binary-koan, line 17:
```suggestion
```

package.json
```diff
@@ -35,20 +34,17 @@
...
+    "test:e2e:headed": "bash scripts/test.sh playwright test --headed",
     "typecheck": "pnpm run with-dev-env pnpm run build && pnpm run-on-packages typecheck && pnpm run with-dev-env tsc --noEmit && pnpm run db:check",
     "services:start": "pnpm with-dev-env pnpm supabase start && (docker ps | grep helperai-nginx || pnpm nginx:start)",
     "services:stop": "pnpm supabase stop && docker stop helperai-nginx || true",
     "nginx:start": "docker run -d --rm --name helperai-nginx -p 80:80 -p 443:443 --add-host host.docker.internal:host-gateway -v \"$(pwd)/scripts/docker/local-nginx/helperai_dev.conf:/etc/nginx/conf.d/default.conf\" -v \"$(pwd)/scripts/docker/local-nginx/certs:/etc/ssl/certs\" nginx:1.27.5",
     "generate-ssl-certificates": "bash scripts/generate-ssl-certificates.sh",
     "run-on-packages": "pnpm run --filter=!helperai --if-present --workspace-concurrency=1",
-    "with-dev-env": "dotenv -e ./.env.development.local -e ./.env.local --"
+    "with-dev-env": "dotenv -e ./.env.development.local -e ./.env.local --",
+    "with-test-env": "dotenv -e ./.env.test.local --"
```
binary-koan, line 47:
```suggestion
    "with-dev-env": "dotenv -e ./.env.development -e ./.env.development.local -e ./.env.local --",
    "with-test-env": "dotenv -e ./.env.test -e ./.env.test.local --"
```

I think we need these?

scripts/test.sh
```diff
@@ -0,0 +1,108 @@
...
+    echo "📁 Files in current directory:"
+    ls -la .env* 2>/dev/null || echo "No .env files found"
+    exit 1
+else
+    echo "✅ .env.test.local found"
+fi
+
+# Source the environment file to get SUPABASE_PROJECT_ID
+set -o allexport
+source .env.test.local
```
binary-koan, line 55:
```suggestion
source .env.test
source .env.test.local
```

scripts/test.sh
```diff
@@ -0,0 +1,108 @@
...
+set +o allexport
+
+# Check for existing Supabase containers and clean them up if found
+echo "🔍 Checking for existing Supabase containers for project ${SUPABASE_PROJECT_ID}..."
```
binary-koan, line 59: I have a feeling some of this isn't needed; will merge for now but worth cleaning up when we've figured out db resetting

binary-koan (commented):
Added some suggestions (I think we can remove the vars from the sample files once we load .env.development and .env.test correctly?)

Could you also rename `scripts/test.sh` to `scripts/e2e.sh` to avoid confusion with unit tests?

Looks good apart from those!

dixitt5:
@binary-koan have made the suggested changes, please have a look.

binary-koan:
LGTM! Just moved another couple of variables to .env.test so you don't need to fill them in explicitly

dixitt5:
@binary-koan looks like the supabase preview run is not getting the right env variables, let me also check if there's something I can quick fix.

Issue #741, opened 2025-07-11
Screenshot feature doesn't work


Hanaffi:
#### Scenario:
Open helper widget
Write screenshot
Check that checkbox and click send

<img width="1920" height="1005" alt="Image" src="https://github.com/user-attachments/assets/d553399b-501a-4604-bdaa-ffca225515e6" />

chuporustum:
This will be fixed as part of this PR - https://github.com/antiwork/helper/pull/737

https://github.com/user-attachments/assets/b50b871c-4fa4-477d-b163-ad820ce6eea1

chuporustum:
> [@chuporustum](https://github.com/chuporustum) I don't think your PR fixes this issue. I can already send screenshots locally but when it comes to other websites it breaks.

oh I see. You are right. My bad for the false alarm

PR #744, opened 2025-07-12
fix: remove screenshot CORS

Hanaffi:
Resolves: https://github.com/antiwork/helper/issues/741

We face CORS issue when we try to load screenshot worker from different origin.
image
which prevents us from submitting a screenshot.
image

### Solutions:
1- Allow CORS on server: This isn't a good solution because we allow people to self-host and then they have to remember to allow CORS on their server too.
2- Inline the worker code into the bundle 

## Summary by CodeRabbit

* **Refactor**
  * Improved screenshot feature by optimizing how the worker script is loaded and initialized, enhancing performance and reliability.

### DIFF
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -21,7 +21,6 @@ import {
 // Import worker code as string
 const workerCode = require("modern-screenshot/dist/worker.js");
 
-// Function to create inline worker URL using Blob
 function createInlineWorkerUrl(): string {
   const blob = new Blob([workerCode], { type: "application/javascript" });
   return URL.createObjectURL(blob);
@@ -774,6 +773,7 @@ class HelperWidget {
       const screenshot = await domToPng(this.screenshotContext);
       this.sendMessageToEmbed({ action: "SCREENSHOT", content: screenshot });
     } catch (error) {
+      // eslint-disable-next-line no-console
       console.error("Failed to take screenshot:", error);
       this.sendMessageToEmbed({ action: "SCREENSHOT", content: null });
     }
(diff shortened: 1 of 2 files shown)


packages/sdk/src/index.ts
```diff
@@ -18,16 +18,20 @@ import {
   type NotificationStatus,
 } from "./utils";
 
+// Import worker code as string
```
slavingia, line 21: No need imo

packages/sdk/src/index.ts
```diff
@@ -755,13 +758,17 @@ class HelperWidget {
...
 
   private async takeScreenshot(): Promise<void> {
-    const { domToPng, createContext } = await import("modern-screenshot");
-    this.screenshotContext ??= await createContext(document.body, {
-      workerUrl: screenshotWorkerUrl,
-      workerNumber: 1,
-      filter: (node) => !(node instanceof HTMLElement && node.className.startsWith("helper-widget")),
-    });
     try {
+      const { domToPng, createContext } = await import("modern-screenshot");
```
binary-koan, line 762: I'm a little confused why bundle splitting here works but loading the worker doesn't - are browsers stricter about worker script origins than page scripts?

packages/sdk/webpack.sdk.cjs
```diff
@@ -31,7 +31,7 @@ module.exports = (env) => {
         },
         {
           test: /modern-screenshot\/dist\/worker.js$/,
-          type: "asset/resource",
+          use: "raw-loader",
```
binary-koan, line 34: This is deprecated, we should use a different [asset module](https://webpack.js.org/guides/asset-modules/)

binary-koan (commented):
Added a couple of comments, could you check?

binary-koan (approved):
LGTM!

PR #748, opened 2025-07-13
feat: added save/edit actions

Adarsh9977:
As discussed with @slavingia in https://github.com/antiwork/helper/pull/564

- [x] Added new edit action for edit.
- [x] removed the saving Indicator for each updation.
- [x] Removed multiple API calls for each fields to update.

https://github.com/user-attachments/assets/2b390aab-006d-41eb-af72-e25cc04887b9

### DIFF
--- a/app/(dashboard)/settings/team/teamMemberRow.tsx
+++ b/app/(dashboard)/settings/team/teamMemberRow.tsx
@@ -1,16 +1,14 @@
 "use client";
 
-import { Trash } from "lucide-react";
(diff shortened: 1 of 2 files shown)


Adarsh9977:
When user is not admin

Screenshot 2025-07-14 at 12 18 35 AM

binary-koan:
In general we want live-updating fields rather than edit/save buttons (across all settings), so I'd rather keep the old design and saving indicator. Very open to removing those multiple API calls though if you don't mind updating to just that?

slavingia:
Agree on live updating

PR #798, opened 2025-07-21
feat: add a CI action for running e2e tests

dixitt5:
### Environment Configuration Updates:
* Added a new optional `HELPER_HOST` environment variable in `lib/env.ts` to allow overriding the default Helper widget host URL (`https://helperai.dev`). This is reflected in `.env.local.sample` and `.env.test` files. [[1]](diffhunk://#diff-8e80df407bc3dc35ee87c4b832b6a879263808697d8554a9039ce1debe6929e7R63-R65) [[2]](diffhunk://#diff-aa29fbbd3836acca1f42a768c116ec7f0e8c1c8a4b9b4a9e73059100aa71252cR23-R26) [[3]](diffhunk://#diff-bd8c5a5839b43575d62ee4f8b0cf6688f479ddd92251d43708b17259a48997baL12-R22)

### E2E Testing Enhancements:
* Introduced a new GitHub Actions workflow (`.github/workflows/e2e.yml`) to automate Playwright E2E tests for pull requests and pushes to the `main` branch.
* Refactored `scripts/e2e.sh` to simplify the E2E test execution process, removing redundant setup logic and relying on pre-configured environments. For setup, we already had `setup-e2e-tests.sh` file, so updated that just runs the e2e setup. 
* Updated `playwright.config.ts` to conditionally open the HTML report based on the CI environment.

### Codebase Refactoring:
* Replaced hardcoded `https://helperai.dev` URLs with the `HELPER_HOST` environment variable in multiple files, including widget test pages (`page.tsx`) and utility functions (`constants.ts`).

### Additional Changes:
* Added a new script `test:e2e:setup` in `package.json` to initialize the E2E testing environment.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the Helper widget host URL via the `HELPER_HOST` environment variable, allowing for custom or local host overrides.
  * Introduced mock OpenAI models for testing and CI environments to simulate API behavior without real network calls.

* **Improvements**
  * Updated various pages and logic to use the `HELPER_HOST` environment variable instead of hardcoded URLs.
  * Enhanced E2E testing setup and execution with new scripts and a GitHub Actions workflow for automated Playwright tests.
  * Improved Playwright test report handling based on environment (CI vs. local).
  * Increased login redirect timeout in E2E auth setup for greater reliability.

* **Documentation**
  * Rewrote E2E test README to reflect new automated setup and streamlined test-running procedures.

* **Chores**
  * Updated environment sample and test files to document and utilize the new `HELPER_HOST` variable.
  * Added and refactored scripts for more robust E2E test environment management.


dixitt5:
@binary-koan Apologies for the delayed response!

I see two possible approaches here:

1. **Mock OpenAI calls in widget tests** – This would work but for e2e tests, it would require mocking the API call itself, which feels a bit off, since it wouldn't truly reflect the production behavior.

2. **Restrict the action trigger** – Either enable the action only on merges to `main`, or follow your suggestion to trigger it only for PRs created from branches within the origin repo. This seems cleaner and avoids false positives from forks or external contributors.

Happy to go with the second approach if you agree, it feels like the more reliable long-term solution.

slavingia:
However, the test should not be hitting OpenAI. We should be mocking those.

binary-koan:
Can we have an env var that switches the language models to test ones maybe? https://ai-sdk.dev/docs/ai-sdk-core/testing

Still no easy way to control it from Playwright but just replying with the same string as the prompt (for example) might be enough for our current suite, and we can worry about more advanced control if we add tests that really need it

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,71 @@
+name: Playwright E2E Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
slavingia, line 7:
```suggestion
on:
  push:
```

Issue #808, opened 2025-07-23
Supabase deployment issue when resetting db


IrishWhiskey:
### What

I am following the setup guide step-by-step, but I run into issues when resetting the database for the first time. 
The `pnpm db:reset` command fails, initially due to missing ssl certificates, and then because of other issues. After running `pnpm generate-ssl-certificates`, and again `pnpm db:reset` I get:
```
All tables are empty. Starting seed process...
Database seed failed: Error: Failed to create user: support@gumroad.com
    at assert (/root/dev/helper/components/utils/assert.ts:3:11)
    at assertDefined (/root/dev/helper/components/utils/assert.ts:8:3)
    at <anonymous> (/root/dev/helper/db/seeds/seedDatabase.ts:71:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Promise.all (index 0)
    at async seedDatabase (/root/dev/helper/db/seeds/seedDatabase.ts:60:19)
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Command failed with exit code 1.
```

This has been executed on a Digital Ocean droplet with 16Gb of RAM and 4 AMD cores, and starting image "Docker on Ubuntu 22.04".

### Why
Looking at the logs of `helper-nginx` it seems like a connection timed out error
```
root@antiwork-helper:~/dev/helper# docker logs helperai-nginx
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2025/07/20 23:10:02 [notice] 1#1: using the "epoll" event method
[...]

PR #812, opened 2025-07-24
fix: issues with local database deployment

IrishWhiskey:
This PR addresses several issues encountered during local deployment of the database on Ubuntu/Linux, including errors with NGINX proxying, SSL certificates, and duplicate key violations in seeding. The error has been documented in issue #808.

All fixes have been tested locally, resolving the original `pnpm db:reset` failures. Below is a summary for each one:

## Generate SSL Certificates in db:reset Command
Local HTTPS proxying via NGINX required self-signed certificates, but they weren't generated automatically, leading to connection refused errors or container startup failures.
The fix was adding a step in the `db:reset` script to generate the certificates before starting services. This leverages the same logic contained inside the `pnpm dev` command. The documentation has been updated accordingly.

## Change proxy_pass to Use Internal Supabase Kong HTTP Port

The original `proxy_pass` (`https://host.docker.internal:54321`) in NGINX config caused 502 Bad Gateway errors due to  `host.docker.internal` resolution issues.
The PR fixes this by updating `scripts/docker/local-nginx/helperai_dev.conf` to proxy to Supabase's internal Kong container on HTTP port `8000` (`proxy_pass http://supabase_kong_helper:8000;`). Also, update `nginx:start` to explicitely join the Supabase network (`--network supabase_network_helper`).

## Ensure User Doesn't Exist Before Seeding into DB
The seed script attempted to create a user (`support@gumroad.com`) that already existed in Supabase auth, causing errors.
This was fixed by modifying `db/seeds/seedDatabase.ts` to check for existing users with `supabase.auth.admin.listUsers()` before creation, skipping if already present.

## Summary by CodeRabbit

* **New Features**
  * Added automated SSL certificate generation and management for local development and database reset processes.
  * Introduced a new script to streamline SSL certificate setup across different operating systems.

* **Bug Fixes**
  * Improved user creation during database seeding to prevent duplicate users.

* **Documentation**
  * Updated development documentation to clarify SSL certificate generation steps.

* **Chores**
  * Updated scripts and Docker configuration for improved local development workflow and network setup.


tests/support/factories/platformCustomers.ts
```diff
@@ -14,6 +15,12 @@ export const platformCustomerFactory = {
         links: { Impersonate: faker.internet.url() },
         ...overrides,
       })
+      .onConflictDoUpdate({
```
binary-koan, line 18: Did you get test errors that required this? The db is supposed to be completely cleared between tests so worth tracking down what's wrong rather than adding this

package.json
```diff
@@ -40,7 +40,8 @@
     "typecheck": "pnpm run with-dev-env pnpm run build && pnpm run-on-packages typecheck && pnpm run with-dev-env tsc --noEmit && pnpm run db:check",
     "services:start": "pnpm with-dev-env pnpm supabase start && (docker ps | grep helperai-nginx || pnpm nginx:start)",
     "services:stop": "pnpm supabase stop && docker stop helperai-nginx || true",
-    "nginx:start": "docker run -d --rm --name helperai-nginx -p 80:80 -p 443:443 --add-host host.docker.internal:host-gateway -v \"$(pwd)/scripts/docker/local-nginx/helperai_dev.conf:/etc/nginx/conf.d/default.conf\" -v \"$(pwd)/scripts/docker/local-nginx/certs:/etc/ssl/certs\" nginx:1.27.5",
+    "nginx:start": "docker run -d --rm --name helperai-nginx -p 80:80 -p 443:443 --add-host host.docker.internal:host-gateway --network supabase_network_helper -v \"$(pwd)/scripts/docker/local-nginx/helperai_dev.conf:/etc/nginx/conf.d/default.conf\" -v \"$(pwd)/scripts/docker/local-nginx/certs:/etc/ssl/certs\" nginx:1.27.5",
```
binary-koan, line 43: Do we still need the `--add-host` if `host.docker.internal` doesn't work anyway?

binary-koan (commented):
Thanks for digging into this! Added a few comments/questions

tests/support/factories/platformCustomers.ts
IrishWhiskey: Actually, I double-checked one more time and this is needed not to make `db:reset` log an error. The `platformCustomerFactory` object is being referenced from the `seedDatabase.ts` file, when creating a customer from  the emails in the `conversationRecords`. But there are entries with duplicate email addresses (i.e. id 21 and 29).

binary-koan (approved):
Looks great, thanks!

Issue #584, opened 2025-06-24
Add E2E tests with Playwright


binary-koan:
- [ ] Add Playwright config and example tests
- [ ] Use a separate database for test
- [ ] Add CI action
- [ ] Get basic coverage on all pages

### Why

It's hard to manually test everything with every change and AI should make it easy to keep tests up to date (cursor rules to tell it to). Also makes it easier to take OSS contributions seriously.

uddhavphatak:
Is this up for grabs?

uddhavphatak:
Yeah yeah it's fine, you can work on it

binary-koan:
I haven’t seen any updates on it - @dixitt5 @uddhavphatak how are you getting on?

IMO getting a separate test database set up is the most pressing thing that will make everything else easier (although it may also be one of the trickier things if Supabase doesn’t have an easy solution)

sankalp1999:
I think #726 fixes broken E2E. I can make a separate PR for the rest?

Anshgrover23:
@dixitt5 can i help you in solving other parts of this issue or do you need any help with? I am eager to do this, if u are stuck on anything?

binary-koan:
> do you think using supabase branching would be a good idea to solve "Use a separate database for test" ?

@Hanaffi missed this one sorry. Looks like local Supabase doesn't support branching? If there is a way to get multiple databases on a local Supabase setup it would be nicer than the current approach of running two though

slavingia:
> Reset database between tests

Maybe it's fine not to do this. I notice this is the standard in Rails land but with front-end stuff it's kind of nice to be able to write tests that already have preexisting data for realism.

slavingia:
True though it doesn't seem to be an issue in the other repos we do this with (though they're still very small). Adding and improving an invoice for example - unlikely to become flakey due to other data

PR #831, opened 2025-08-01
test: add team settings E2E tests

Udit-takkar:
Address Team settings (invite, edit/remove member, admin vs member permissions) in https://github.com/antiwork/helper/issues/584

Screenshot 2025-08-01 at 5 33 41 AM

### DIFF
--- a/tests/e2e/utils/page-objects/teamSettingsPage.ts
+++ b/tests/e2e/utils/page-objects/teamSettingsPage.ts
@@ -1,0 +1,193 @@
+import { expect, Page } from "@playwright/test";
+import { generateTestEmail } from "../test-helpers";
+import { BasePage } from "./basePage";
+
+/**
+ * Page object for Team Settings functionality.
+ *
+ * Note: Many operations (invite, remove, role changes) require admin permissions.
+ * The test suite uses support@gumroad.com which is created with admin role during seeding.
+ */
+export class TeamSettingsPage extends BasePage {
+  constructor(page: Page) {
+    super(page);
+  }
+
+  async navigateToTeamSettings() {
+    await this.goto("/settings/team");
+    await this.waitForPageLoad();
+  }
+
+  async expectTeamSettingsPage() {
+    await expect(this.page).toHaveURL(/.*settings\/team.*/);
+    await this.expectVisible('[data-testid="team-settings-page"]');
+  }
+
+  async inviteMember(email?: string): Promise<string> {
+    const testEmail = email || generateTestEmail();
+
+    // Fill email input
+    const emailInput = this.page.getByTestId("invite-email-input");
+    await expect(emailInput).toBeVisible();
+    await emailInput.fill(testEmail);
+
+    // Fill name input (required)
+    const nameInput = this.page.getByTestId("invite-name-input");
+    await expect(nameInput).toBeVisible();
+    await nameInput.fill(`Test User ${Date.now()}`);
+
+    // Select permissions
+    const permissionsSelector = this.page.getByTestId("member-role-selector");
+    await expect(permissionsSelector).toBeVisible();
+    await permissionsSelector.click();
+
+    const memberOption = this.page.getByTestId("role-option-member");
+    await expect(memberOption).toBeVisible();
+    await memberOption.click();
+
+    // Submit invitation
+    const submitButton = this.page.getByTestId("invite-member-button");
+    await expect(submitButton).toBeVisible();
+    await submitButton.click();
+
+    return testEmail;
+  }
+
+  async expectMemberInvited(email: string) {
+    // Wait for success toast message
+    const successToast = this.page.locator('[data-sonner-toast]:has-text("Team member added")');
+    await expect(successToast).toBeVisible({ timeout: 10000 });
+  }
+
+  async expectMemberInList(email: string) {
+    const memberRow = this.page.getByTestId("member-row").filter({ hasText: email });
+    await expect(memberRow).toBeVisible();
+  }
+
+  async removeMember(email: string) {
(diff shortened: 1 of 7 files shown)


Udit-takkar:
> @Udit-takkar Thanks for picking this up! I would rather avoid data-testid unless we really need it and test for the actual messages etc - could you try to remove as many as feasible? (just added a guideline in https://github.com/antiwork/helper/issues/584)

@binary-koan  Okay I can do that but using data-testid is generally preffered more as it won't break if you update any text and very easier to find the element while locating instead of adding too many selector class which is not readable and easily understood.

binary-koan:
> it won't break if you update any text

I understand that's the tradeoff, but I think AI tooling can mitigate that and I would rather have tests that simulate user interaction more realistically.

> easier to find the element while locating instead of adding too many selector class which is not readable

Yeah I agree with the sentiment, if the selector gets much more complex than an element type + expected text I'm more open to using it.

IMO `conversations.spec.ts`  and `conversationsPage.ts` are on the right track, mostly finding via text but using testid sometimes when it's significantly easier/clearer. I'm not keen on some of the class selectors (e.g. `mark.bg-secondary-200`) - those would be better with `testid` too as they don't indicate what the element is - but I think finding buttons by text and inputs by placeholder/label is a reasonable balance.

tests/e2e/utils/page-objects/teamSettingsPage.ts
```diff
@@ -0,0 +1,250 @@
...
+    await emailInput.fill(email);
+    
+    await expect(nameInput).toBeVisible();
+    await nameInput.fill(name);
+  }
+
+  async fillInvalidEmail(email: string) {
+    const emailInput = this.page.locator(SELECTORS.EMAIL_INPUT);
+    await emailInput.fill(email);
+  }
```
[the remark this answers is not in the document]
Udit-takkar: The id for email address was already present so i just used that

tests/e2e/utils/page-objects/teamSettingsPage.ts
```diff
@@ -0,0 +1,250 @@
...
+import { BasePage } from "./basePage";
+
+const TIMEOUTS = {
+  ELEMENT_VISIBLE: 10000,
+  NETWORK_IDLE: "networkidle" as const,
+  FORM_STABILITY: 2000,
+} as const;
+
+const SELECTORS = {
+  EMAIL_INPUT: "#email-input",
```
binary-koan, line 12: Please make these `private readonly` variables for consistency with the other page objects

tests/e2e/utils/page-objects/teamSettingsPage.ts
```diff
@@ -0,0 +1,250 @@
...
+  NAME_INPUT: "#display-name-input",
```
binary-koan, line 13: My point was to use things like `input[placeholder="Name"]` for things like this too. Hopefully there's a nice semantic selector that we can use for the selects as well, although not sure off the top of my head (`[role="combobox"]:has-text(...)` or something?)

tests/e2e/utils/page-objects/teamSettingsPage.ts
```diff
@@ -0,0 +1,250 @@
...
+  INVITE_FORM: "form#invite-member-form",
```
binary-koan, line 14: We shouldn't need to test the form existence, we can infer that it's the right form if the field labels and interaction matches what we expect

tests/e2e/utils/page-objects/teamSettingsPage.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  async expectMemberInvited(email: string) {
+    await this.expectToast(MESSAGES.MEMBER_ADDED);
+    await this.expectMemberInList(email);
+  }
+
+  async expectMemberInList(email: string) {
+    const memberRow = this.getMemberRow(email);
+    await expect(memberRow).toBeVisible({ timeout: TIMEOUTS.ELEMENT_VISIBLE });
```
binary-koan, line 64: Is there a way to increase the default timeout instead of specifying this everywhere?

tests/e2e/utils/db-cleanup.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function cleanupTestMembers(emails: string[] = []) {
+  try {
+    if (emails.length === 0) {
+      return;
+    }
+
+    console.log(`Cleaning up ${emails.length} test members:`, emails);
+
+    // Delete specific test members by email
+    for (const email of emails) {
```
binary-koan, line 19: I think it's fine to do this as one `db.delete(...).where(inArray(users.email, ...)` query

tests/e2e/utils/db-cleanup.ts
```diff
@@ -0,0 +1,37 @@
+import { eq, like } from "drizzle-orm";
```
binary-koan, line 1: No change needed in this PR, but maybe separately we can merge `delete-saved-replies.ts` into this and use it similarly in the saved replies tests - I much prefer this approach over a manual command

binary-koan (commented):
This is going in the right direction, added some more comments

tests/e2e/utils/page-objects/teamSettingsPage.ts
Udit-takkar: yeah just a second. i am trying to replace all the selectors

Udit-takkar:
@binary-koan Done. refactored the code to completely remove data-testid and id usage

binary-koan (approved):
Pushed a small UX fix + simplified the page object a bit to follow conventions. Apart from that LGTM, will merge once tests pass!

PR #838, opened 2025-08-01
tests: add E2E tests for customer settings page

Udit-takkar:
This PR add E2E tests for "Customer Settings Page" as mentioned in https://github.com/antiwork/helper/issues/584


Udit-takkar:
> I realised I just merged a different PR for the same page 😅 https://github.com/antiwork/helper/pull/854
Could you look through that and see if you want to improve any of the tests based on your work here?

Oh didn't know about this. Now I have resolved conflicts. in my opinion i only added very limited and concise tests that are necessary and should be faster to run as in the future we would have too many tests and CI would slow slow. Also this file doesn't run serially like in the PR that was merged so should be again much faster and concise

Udit-takkar:
Screenshot 2025-08-07 at 6 09 29 AM

@binary-koan Another thing is the customer settings page test on the main branch is causing this error. while this PR is working fine.

PR #845, opened 2025-08-02
test(e2e) : add tests for website learning

achalbajpai:
### What 
PR address #584 sub-point `Website learning (UI smoke test only)`

### Playwright Test Report 
image

### Local Tests
image

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end tests for the Website Learning UI, including form interactions, validation, and website addition flows.
  * Introduced a new page object to streamline and standardize test interactions with the Website Learning feature.
* **Chores**
  * Improved test reliability by adding a test identifier to website list items.
  * Enhanced toast notification detection with a new utility function and refactored related test code.

### DIFF
--- a/tests/e2e/utils/page-objects/websiteLearningPage.ts
+++ b/tests/e2e/utils/page-objects/websiteLearningPage.ts
@@ -1,6 +1,6 @@
 import { expect, Page } from "@playwright/test";
-import { BasePage } from "./basePage";
 import { waitForToast } from "../toastHelpers";
+import { BasePage } from "./basePage";
 
 interface TestWebsite {
   name: string;
@@ -9,9 +9,10 @@ interface TestWebsite {
 
 export class WebsiteLearningPage extends BasePage {
   private readonly websiteLearningTitle = 'h2:has-text("Website Learning")';
-  private readonly websiteDescription = 'text="Helper will learn about your product by reading your websites to provide better responses."';
+  private readonly websiteDescription =
+    'text="Helper will learn about your product by reading your websites to provide better responses."';
   private readonly addWebsiteButton = 'button:has-text("Add website")';
-  private readonly urlInput = 'input#url';
+  private readonly urlInput = "input#url";
   private readonly urlLabel = 'label[for="url"]';
   private readonly cancelButton = 'button:has-text("Cancel")';
   private readonly submitButton = 'form button[type="submit"]';
@@ -84,7 +85,7 @@ export class WebsiteLearningPage extends BasePage {
     const websiteItem = this.page.locator(this.websiteItems).filter({
       has: this.page.locator(`text="${websiteName}"`),
     });
-    
+
     await expect(websiteItem).toBeVisible({ timeout: 15000 });
   }
 }
--- a/tests/e2e/utils/page-objects/websiteLearningPage.ts
+++ b/tests/e2e/utils/page-objects/websiteLearningPage.ts
@@ -1,0 +1,113 @@
+import { expect, Page } from "@playwright/test";
+import { BasePage } from "./basePage";
+
+interface TestWebsite {
+  name: string;
+  url: string;
+}
+
+export class WebsiteLearningPage extends BasePage {
+  private readonly websiteLearningTitle = 'h2:has-text("Website Learning")';
+  private readonly websiteDescription = 'text="Helper will learn about your product by reading your websites to provide better responses."';
+  private readonly addWebsiteButton = 'button:has-text("Add website")';
+  private readonly urlInput = 'input#url';
+  private readonly urlLabel = 'label[for="url"]';
+  private readonly cancelButton = 'button:has-text("Cancel")';
+  private readonly submitButton = 'form button[type="submit"]';
+  private readonly errorText = '.text-sm.text-destructive';
+  private readonly websiteItems = 'div.rounded-lg.border';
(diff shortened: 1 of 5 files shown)


tests/e2e/utils/page-objects/websiteLearningPage.ts
```diff
@@ -0,0 +1,113 @@
...
+      if (await toast.count() > 0) {
+        await expect(toast.first()).toBeVisible({ timeout: 8000 });
+        found = true;
+        break;
+      }
+    }
+    
+    if (!found) {
+      const fallbackToast = this.page.locator(`text=${message}`).first();
+      await expect(fallbackToast).toBeVisible({ timeout: 8000 });
```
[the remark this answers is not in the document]
achalbajpai: I don't think this is needed in our case

tests/e2e/website-learning/websiteLearning.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  test("displays main website learning interface", async () => {
+    await websiteLearningPage.expectWebsiteLearningSection();
+    await websiteLearningPage.expectAddWebsiteButton();
+  });
+
+  test("shows add website form on button click", async () => {
+    await websiteLearningPage.clickAddWebsite();
+    await websiteLearningPage.expectAddWebsiteForm();
+  });
```
binary-koan, line 29: I think we can merge these

```suggestion
  test("displays the website learning section and add website form", async () => {
    await websiteLearningPage.expectWebsiteLearningSection();
    await websiteLearningPage.expectAddWebsiteButton();

    await websiteLearningPage.clickAddWebsite();
    await websiteLearningPage.expectAddWebsiteForm();
  });
```

tests/e2e/utils/page-objects/websiteLearningPage.ts
```diff
@@ -0,0 +1,113 @@
...
+export class WebsiteLearningPage extends BasePage {
+  private readonly websiteLearningTitle = 'h2:has-text("Website Learning")';
+  private readonly websiteDescription = 'text="Helper will learn about your product by reading your websites to provide better responses."';
+  private readonly addWebsiteButton = 'button:has-text("Add website")';
+  private readonly urlInput = 'input#url';
+  private readonly urlLabel = 'label[for="url"]';
+  private readonly cancelButton = 'button:has-text("Cancel")';
+  private readonly submitButton = 'form button[type="submit"]';
+  private readonly errorText = '.text-sm.text-destructive';
+  private readonly websiteItems = 'div.rounded-lg.border';
```
binary-koan, line 18: I think we can improve these a bit

* For the error text we don't care about where it is and what class name it has, so I think we can just check the page contains the expected error text anywhere
* For the website items I think it's worth adding a `data-testid`, this class is too generic and doesn't indicate what the element actually is

tests/e2e/utils/page-objects/websiteLearningPage.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  generateTestWebsite(): TestWebsite {
+    const timestamp = Date.now();
+    const url = `https://test-${timestamp}.example.com`;
+    const name = `test-${timestamp}.example.com`;
+    return { name, url };
+  }
+
+  async expectToastMessage(message: string) {
```
binary-koan, line 81: Can we extract this to somewhere shared? (I see we have a similar `waitForToast` function in the saved replies page, and I expect we'll want it in a bunch of other places too)

binary-koan (commented):
Thanks! Added a few comments

achalbajpai:
> Thanks! Added a few comments

Thanks for the review . I have addressed all the comments mentioned. I hope the PR is good to be merged, I’m happy to fix any other issues if there are any left.

image

tests/e2e/utils/toastHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+  const toastSelectors = [
+    '[data-sonner-toast]',
+    '[data-sonner-toaster] li',
+    '.toaster li',
+    '[role="status"]',
+    '[role="alert"]',
+    '[aria-live="polite"]'
+  ];
+  
+  for (const selector of toastSelectors) {
```
binary-koan, line 13: We need to keep the `toastFound` logic here, otherwise this may be super slow while it waits for all the possible selectors. Alternatively we could just use a single selector instead of an array - all our toasts are the same so I'm not sure why we need to test a bunch of different selectors 🤔

binary-koan (commented):
Added one more comment, looks good apart from that!

tests/e2e/utils/toastHelpers.ts
achalbajpai: > We need to keep the `toastFound` logic here, otherwise this may be super slow while it waits for all the possible selectors. Alternatively we could just use a single selector instead of an array - all our toasts are the same so I'm not sure why we need to test a bunch of different selectors 🤔

Added back the toastFound function; it does improve the speed. Also, since all the selectors were finding the same element, it’s better to keep just one , rather than covering all scenerios.

achalbajpai:
> Added one more comment, looks good apart from that!

Addressed the issues mentioned @binary-koan  , I hope all the changes are good to go.

binary-koan (approved):
LGTM!

PR #868, opened 2025-08-06
fix: update devIndicators configuration for test environment

dixitt5:
### Summary
The Next.js dev indicator in the widget was interfering with Playwright tests by overlapping the "Take Screenshot" button, causing test failures.
    → It is now disabled in test environments to prevent such issues.
    → The indicator will still appear in case of runtime errors but won't be shown by default during tests.

- Verified the fix by running the same CI workflow on my forked repo.
    ✅ The test suite now passes successfully.
    
[workflow on my forked repo](https://github.com/dixitt5/helper/actions/runs/16767999546/job/47476963517)

### Screenshot:
image

### Dev indicator displayed on localhost:
image

## Summary by CodeRabbit

* **Chores**
  * Disabled development indicators in the application configuration based on the test environment setting.


binary-koan (approved):
Good call, thanks!

PR #882, opened 2025-08-07
Upgrade to AI SDK 5

binary-koan:


binary-koan:
This upgrade is a massive pain and we should be able to use GPT-5 without it, so will delay for now

PR #891, opened 2025-08-09
Cache client tools for admin UI 

techmannih:
Fixes #890 

## Summary by CodeRabbit

* New Features
  * Added optional “customer-specific tools” flag across API, client SDK, and React hook; propagated to auto-response events.
  * Introduced server-side caching of server-backed tools per customer email with a global fallback.
  * Tool listings now include cached customer tools alongside existing tools.

* Chores
  * Added database table to store cached tools and exposed schema via index.
  * Background caching runs asynchronously after chat requests.
  * Updated event payload schema to accept the customer-specific tools flag.

* Tests
  * Added tests for auto-response payload forwarding and caching behaviors.


techmannih:
@binary-koan Could you please review it? I'm still working on the end-to-end tests.

techmannih:
@binary-koan Do I need to do ll suggestion given by coderabbitai bot ?

binary-koan:
Some of them; I've resolved the ones that don't look relevant/useful

db/schema/cachedClientTools.ts
```diff
@@ -0,0 +1,20 @@
...
+import { withTimestamps } from "../lib/with-timestamps";
+import { ToolRequestBody } from "@helperai/client";
+
+export const cachedClientTools = pgTable(
+  "cached_client_tools",
+  {
+    ...withTimestamps,
+    id: bigint({ mode: "number" }).primaryKey().generatedByDefaultAsIdentity(),
+    customerEmail: text(),
+    platformCustomerId: bigint("platform_customer_id", { mode: "number" }),
```
binary-koan, line 11: We shouldn't have both of these - ideally we should always link to the platform customer (e.g. find or create as part of the caching?) but if that doesn't work well we might need to switch to email instead

trpc/router/mailbox/conversations/tools.ts
```diff
@@ -41,15 +42,35 @@ export const toolsRouter = {
       }
     });
 
+    const cachedTools = await getCachedClientTools(conversation.emailFrom);
+    const cachedToolList = cachedTools
+      ? Object.entries(cachedTools).map(([name, tool]) => ({
+          name,
```
binary-koan, line 48: Maybe we can `startCase` this to make it nicer to read

binary-koan (commented):
Added another couple of comments while I'm here

app/api/chat/conversation/[slug]/message/route.ts
```diff
@@ -66,12 +71,16 @@ export const POST = withWidgetAuth<{ slug: string }>(async ({ request, context:
...
+  try {
+    await triggerEvent(
+      "conversations/auto-response.create",
+      { messageId: userMessage.id, tools },
+      { sleepSeconds: 5 * 60 },
+    );
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error("Failed to enqueue auto-response", err);
+  }
```
Comment (binary-koan) [inline app/api/chat/conversation/[slug]/message/route.ts:83]: Generally speaking we can just let things like this fail (did you run into problems with it?)

app/api/chat/conversation/[slug]/message/route.ts
```diff
@@ -14,7 +14,12 @@ export const OPTIONS = () => corsOptions("POST");
 
 export const POST = withWidgetAuth<{ slug: string }>(async ({ request, context: { params } }, { session }) => {
   const { slug } = await params;
-  const { content, attachments = [], tools } = createMessageBodySchema.parse(await request.json());
+  const {
+    content,
+    attachments = [],
+    tools,
+    customerSpecificTools,
```
Comment (binary-koan) [inline app/api/chat/conversation/[slug]/message/route.ts:21]: Currently unused? (shouldn't it be passed to `conversations/auto-response.create`?)

Would be good to include tests for this route to catch that

binary-koan (commented):
Thanks! Had a more detailed look through and added some comments

binary-koan (approved):
Pushed a few more fixes, LGTM!

Issue #899, opened 2025-08-10
Change date of conversation in list to last message date


Hanaffi:
### What
Change date of conversion from created_at date to the date of last message

<img width="620" height="687" alt="Image" src="https://github.com/user-attachments/assets/5ba9d5ad-0aca-4e9a-9b29-1f73984b9ac6" />

### Why
Its confusing to see the date of conversations to be the date which the conversation was started.
By convention, the date behind a conversation indicates the date of last message (Like Facebook, twitter, and other apps)
This way the order makes sense.

binary-koan:
We also have a longstanding issue where the date shown to admins is the last *user* message, not the last message, which makes it confusing to see which tickets have/haven't been replied to. IMO we should add a `lastMessageAt` column to `conversations` and update it on every new user/staff/assistant message (joining probably isn't performant enough) - then we can use it consistently everywhere

PR #905, opened 2025-08-11
[Feat] Add last message at

Hanaffi:
Resolves: https://github.com/antiwork/helper/issues/899
This PR adds a new column, backfills existing data, and begins writing to it.
A follow-up PR will handle displaying the column’s value.

- [x] Add new column
- [x] Update actions to update this column when we add new message
- [x] Write specs
- [x] Backfill data 

## Summary by CodeRabbit

- **New Features**
  - Conversations now include a “Last message” timestamp that updates for every message and is exposed in serialized conversation output.

- **Chores**
  - Database migration adds the new timestamp column and index.
  - Backfill tooling added to populate existing conversations' last-message timestamps and an npm script to run it.
  - Migration journal updated and minor CONTRIBUTING.md whitespace tweak.

- **Tests**
  - Added test verifying conversations' last-message timestamp is updated.

### DIFF
--- a/db/schema/conversations.ts
+++ b/db/schema/conversations.ts
@@ -72,6 +72,7 @@ export const conversations = pgTable(
     index("conversations_anonymous_session_id_idx").on(table.anonymousSessionId),
     index("conversations_merged_into_id_idx").on(table.mergedIntoId),
     index("conversations_issue_group_id_idx").on(table.issueGroupId),
+    index("conversations_last_message_at_idx").on(table.lastMessageAt),
     index("conversations_conversation_status_last_user_email_created_at_idx")
       .on(table.status, table.lastUserEmailCreatedAt.desc().nullsLast())
       .where(isNull(table.mergedIntoId)),
--- a/db/schema/conversations.ts
+++ b/db/schema/conversations.ts
@@ -26,6 +26,7 @@ export const conversations = pgTable(
     slug: randomSlugField("slug"),
     lastUserEmailCreatedAt: timestamp({ withTimezone: true, mode: "date" }),
     lastReadAt: timestamp({ withTimezone: true, mode: "date" }),
+    lastMessageAt: timestamp("last_message_at", { withTimezone: true, mode: "date" }),
     conversationProvider: text().$type<"gmail" | "helpscout" | "chat">(),
     closedAt: timestamp({ withTimezone: true, mode: "date" }),
     assignedToId: text("assigned_to_clerk_id"),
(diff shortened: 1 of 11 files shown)


db/drizzle/0113_backfill_last_message_at.sql
```diff
@@ -0,0 +1,15 @@
+-- Backfill last_message_at for existing conversations
```
binary-koan, line 1: This will be extremely slow with large numbers of conversations. Best to add something optimised for large datasets in `scripts` ([example](https://github.com/antiwork/helper/blob/v0.5.0/scripts/migrateEncryptedToPlaintext.ts)) and people can optionally run it if they want to backpopulate (note: it's possible to have a conversation with no messages so we need to handle the null case anyway)

lib/data/conversationMessage.ts
```diff
@@ -559,6 +566,15 @@ export const createToolEvent = async ({
     userId,
   });
 
+  await updateConversation(
+    conversationId,
+    {
+      set: { lastMessageAt: new Date() },
```
binary-koan, line 572: I think it's OK not to create these for tool events - they should really be a separate table tbh

scripts/backfillLastMessageAt.ts
```diff
@@ -0,0 +1,64 @@
...
+        WHERE c."last_message_at" IS NULL
+          AND m."status" != 'draft'
+        GROUP BY c."id"
+        ORDER BY c."id"
+        LIMIT ${BATCH_SIZE}
+      )
+      UPDATE "conversations_conversation" AS c
+      SET "last_message_at" = candidates.last_message_at
+      FROM candidates
+      WHERE c."id" = candidates."id";
```
binary-koan, line 40: Would be nice to add `RETURNING id` and `WHERE id > ${lastId}` since we're sorting by ID - it'll make this much more performant

db/schema/conversations.ts
```diff
@@ -26,6 +26,7 @@ export const conversations = pgTable(
     slug: randomSlugField("slug"),
     lastUserEmailCreatedAt: timestamp({ withTimezone: true, mode: "date" }),
     lastReadAt: timestamp({ withTimezone: true, mode: "date" }),
+    lastMessageAt: timestamp("last_message_at", { withTimezone: true, mode: "date" }),
```
binary-koan, line 29: We should index this since we'll want to sort by it in the UI (didn't notice before sorry)

binary-koan (commented):
Another couple of small comments, looks good otherwise!

binary-koan (approved):
LGTM!

PR #945, opened 2025-08-16
[BUG] Cross chat contamination

Hanaffi:
Resolves: https://github.com/antiwork/helper/issues/930

If you try to send a message in conversation A, you will see the same message and reply in all other conversations :exploding_head: 

### Before
https://www.loom.com/share/9688534261bc4cf38de063eb2e2bb69e
### After
https://www.loom.com/share/f32c4cfb971f44cabf9a66ef7d50e238

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where messages and typing/escalation status could carry over when switching conversations.
  * Conversation views now reset per-conversation state on switch, preventing cross-contamination and ensuring the correct state is shown for each conversation.
  * Improves reliability and clarity when navigating between conversations, with no changes to the public API.

### DIFF
--- a/components/widget/Conversation.tsx
+++ b/components/widget/Conversation.tsx
@@ -184,6 +184,7 @@ export default function Conversation({
       setData(undefined);
       setIsAgentTyping(false);
       setMessages([]);
+      setIsEscalated(false);
     }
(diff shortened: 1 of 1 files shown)


slavingia:
Specs missing

PR #998, opened 2025-08-29
Feature: Cache client-provided tools to show to admins

kanhayaKy:
closes #890

AI Disclosure
GitHub Copilot was used for syntax suggestions and quick typing; all logic and code were reviewed manually.

After 

https://github.com/user-attachments/assets/7e1ea99a-9251-42c7-ba30-9d7bb83c18e5

Before
This is a new feature so nothing added here

Tests

image


tests/app/api/chat/conversation/[slug]/message/route.test.ts
```diff
@@ -39,10 +39,17 @@ describe("POST /api/chat/conversation/[slug]/message", () => {
     mockSession = { isAnonymous: false, email: "test@example.com" };
     mockMailbox = mailbox;
 
+    const tools = {
+      testServerTool: {
+        description: "A tool for testing server functionality",
+        parameters: {},
+        serverRequestUrl: "https://example.com/api/tools/testServerTool"
+      }
+    }
```
Comment (kanhayaKy) [inline tests/app/api/chat/conversation/[slug]/message/route.test.ts:48]: To tests the tool caching feature

binary-koan:
Thanks for working on this, but https://github.com/antiwork/helper/pull/1001 is closer to being mergeable so we'll proceed with that one

kanhayaKy:
@binary-koan This PR was raised before the other one, so I’d appreciate if we could review this as well.

kanhayaKy:
@binary-koan  I also noticed that this PR already includes most of the changes you recommended in the other one. wouldn't that make it closer to being mergeable?

CC : @slavingia

kanhayaKy:
@binary-koan Thanks for taking the time to review both PRs. Since neither has been merged yet and both have their own pros and cons, I’d really appreciate the opportunity to improve this one.

Could you share what this PR is currently missing or where it falls short? I’d be happy to address the feedback and rework it accordingly. It would be great if both contributions could be given a fair chance to evolve.

kanhayaKy:
@binary-koan @slavingia 

I’ve reviewed the other PR, and here are a few areas where I believe this one adds more value:

1. **Enhanced Test Widgets** – The updates here improve the visual testing experience, making it easier to validate the new feature and related changes in the future.
2. **Flexible Nested `tool` Field** – This structure allows fields to be added or removed based on `ToolRequestBody` without requiring new migrations, making it easier to maintain over time.
3. **Avoids Redundant Cache Fields** – This PR doesn’t include unsupported properties like `platformCustomerId`, keeping the cache table more focused and maintainable.

That said, I’d really appreciate your input, especially on areas where this PR could be stronger. It’s a great opportunity for me to learn and improve.

binary-koan:
Thanks for the review! I actually disagree with 1 and 2 (see below) but 3 is good point I missed in my last review. tbh I think mashing the PRs together would be the ideal option in this case, so I'm trying to make sure to give you credit where it's due in my comments on the other one.

1. I see your point, but I would rather keep the test widget simple so that it's a template for how to integrate Helper on your own site. I think the feature can be adequately covered by unit tests.
2. I prefer the more explicit database columns because they encourage migrating data cleanly, rather than ending up with multiple possible schemas in a JSON field.

(Another thing I remember noticing here was the snake_case naming of table fields - it's a small fix but to me it stood out as a very avoidable inconsistency that makes me less confident about the rest of the change.)

kanhayaKy:
@binary-koan I noticed that the requirement for this issue is to cache the tools for admins and eventually deprecate the tools API. The other PR caching the tools for the clients , which doesn’t meet the requirement.

I’ve created a new PR to handle this: https://github.com/antiwork/helper/pull/1009 . It caches the tools for admins as intended and the tools api can be easily deprecated when needed. Let me know your thoughts and feedback.

PR #1010, opened 2025-09-03
feat: auto assign replies to self

### LINKED ISSUE #1008: "Auto-assign replies to self" setting
### What

Setting per user to automatically assign tickets to yourself when you reply (if unassigned?)

### Why

Several people do this manually which takes time.

devsargam:
resolves #1008

Before:
When a ticket is unassigned and a user replies to that ticket the ticket remains unassigned

After:
When a ticket is unassigned and a user replies to that the replying user is assigned for that ticket

Demo:
<video src="https://github.com/user-attachments/assets/7c6fbde5-5684-4a7f-b3d4-262776ffcff7" height="300px" />

AI Disclosure
No AI was used to write any code

### DIFF
--- a/app/(dashboard)/[category]/conversation/messageActions.tsx
+++ b/app/(dashboard)/[category]/conversation/messageActions.tsx
@@ -419,7 +419,7 @@ export const MessageActions = () => {
               <Button
                 size={isAboveMd ? "default" : "sm"}
                 variant="outlined"
-                onClick={() => handleSend({ assign: false, close: false })}
+                onClick={() => handleSend({ assign: !conversation?.cc, close: false })}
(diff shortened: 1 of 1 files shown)


devsargam:
Hey @binary-koan implemented a super basic version of if ticket is unassigned just assign the ticket to the user.

Few questions:
- Should we have a field in the settings areas such that this feature can be like a feature flag?
- When reply and close is clicked should the user be assigned in that case too?

Will write the spec files when the implementation is finalized. Thank you.

binary-koan:
Yes, we should have a user preference for it, and it should be on reply and close as well

binary-koan:
(please reopen when done)

Issue #1007, opened 2025-09-02
Fix performance of closed ticket search


binary-koan:
### What

Currently on Gumroad's instance, searching closed tickets is prohibitively slow and often times out. We should make all filters work with reasonable performance (currently at least search by keyword + "replied by" filter are slow).

sm17p:
Can we get a get a few stats like 

1. Size of table
2. Number of rows in those tables 
3. Sizing of indexes in those tables
4. P99 

What is an acceptable P99 for this path?

binary-koan:
It's working well, thanks!

PR #1021, opened 2025-09-04
improve performance for closed tickets search

MuskanPaliwal:
Part of #1007 
### AI Disclosure

Used Claude code to write test scripts for the performance optimization and applying db migrations, rest all the changes are done by me.

### Problem

The agent inbox (conversation list) is rendered and refreshed continuously. Each load applies message-scoped predicates (e.g., “replied by”) and sorts closed tickets. Without targeted indexes, the planner degraded into per-row scans and explicit sorts as data grew, increasing end‑to‑end latency.

### Changes

Introduced two targeted partial B‑Tree indexes, aligned with the query’s correlation keys and ordering.

  * **Closed ordering** (partial; satisfies predicate + `ORDER BY`)

    ```sql
    -- Closed ordering (partial; satisfies predicate + ORDER BY)
    CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_conversations_closed"
    ON "conversations_conversation" ("closed_at" DESC NULLS LAST)
    WHERE ("conversations_conversation"."status" = 'closed'
      AND "conversations_conversation"."merged_into_id" IS NULL);
    ```

  * **Replied-by filter** (partial; correlation-friendly composite)

    ```sql
    -- Replied-by filter (partial; correlation-friendly composite)
    CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_messages_replied_by_filter"
    ON "messages" ("conversation_id", "clerk_user_id")
    WHERE ("messages"."deleted_at" IS NULL AND "messages"."role" = 'staff');
    ```

  * Keyword search remains on the existing GIN index over `messages.search_index` (array containment).

### Why this works (planner mechanics)

  * **Closed ordering**

      * **Before**: filter `status='closed'` → explicit `Sort` on `closed_at DESC NULLS LAST`.
      * **After**: The partial index satisfies both the predicate and the ordering; the planner uses an ordered index scan (no global sort, fewer buffers).

  * **Replied-by `EXISTS` (correlated subquery)**

      * **Shape**: `EXISTS` on `messages` with predicates (`conversation_id`, `role='staff'`, `userId ∈ S`, `[createdAt window]`).
      * **Before**: Bitmap/Seq Scans on `messages` per conversation, then role/user filtering → O(k) row touches per parent conversation.
      * **After**: Composite (`conversation_id`, `clerk_user_id`) + partial predicate enables near O(1) probes per conversation, often an `Index Only Scan` with `Heap Fetches: 0`.

  * **Keyword search**

      * Dominated by GIN; the new B‑Tree indexes are orthogonal, so pure keyword performance is stable (as expected).

### Query plan highlights (salient diffs)

  * **Closed lists:**
      * **Before**: `Sort` on `closed_at`.
      * **After**: `Index Scan` via `"idx_conversations_closed"`; no explicit sort.
  * **RepliedBy `EXISTS`:**
      * **Before**: `Bitmap Heap/Seq Scan` on `"messages"` + role/user filter.
      * **After**: `Index Only Scan` via `"idx_messages_replied_by_filter"`, `Heap Fetches: 0`, `Semi Join` on `conversation_id`.

### Benchmarks (representative)

  * **Large dataset** (\~10k closed; \~11k staff msgs)
      * **Paged (limit 20):**
          * `repliedBy`-only (regression case): \~246 ms → \~24 ms
          * `closed`+`keyword`+`repliedBy`: \~21.8 ms → \~14.6 ms
          * `keyword`-only: \~26–27 ms (GIN‑bound; unchanged)
      * **Full dataset (all pages):**
          * `closed`-only: \~314 ms → \~309 ms (noise/slight improvement)
          * `closed`+`keyword`+`repliedBy`: \~38.0 ms → \~31.9 ms
  * **Small dataset** (\~500 total)
      * All latencies are single‑digit ms with/without indexes; at a tiny scale, index overhead can equal benefits (expected).

### Migration (safe deploy)

Create indexes concurrently to avoid write locks.

```sql
BEGIN;

CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_conversations_closed"
  ON "conversations_conversation" ("closed_at" DESC NULLS LAST)
  WHERE ("status" = 'closed' AND "merged_into_id" IS NULL);

CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_messages_replied_by_filter"
  ON "messages" ("conversation_id", "clerk_user_id")
  WHERE ("deleted_at" IS NULL AND "role" = 'staff');

COMMIT;
```

### Backout

```sql
DROP INDEX CONCURRENTLY IF EXISTS "idx_conversations_closed";
DROP INDEX CONCURRENTLY IF EXISTS "idx_messages_replied_by_filter";
```

### Result

Correlated message filters and closed ticket ordering now map directly to selective partial indexes. The planner replaces costly scans/sorts with ordered scans and targeted index‑only probes, removing the “replied by” regression and producing consistent gains in combined‑filter scenarios, while keeping GIN‑driven keyword performance intact.

All unit tests passing:

Screenshot 2025-09-04 at 6 16 34 PM

Without index performance:
Screenshot 2025-09-04 at 5 34 30 PM

With index performance:
Screenshot 2025-09-04 at 5 34 10 PM


MuskanPaliwal:
@binary-koan , the PR is ready to review.

MuskanPaliwal:
> (Ironically trying to add the indexes in production ran into a timeout, so still working out a way to bypass it and test this)

Indexes are not being used in production then. Can you get the `Production Analyze` output here?

binary-koan:
I'll just merge this (pushed a small tidy-up) and see what happens, we can always drop the indexes again if they don't improve things. Will try to get production ANALYZE output after these are applied!

MuskanPaliwal:
> I'll just merge this (pushed a small tidy-up) and see what happens, we can always drop the indexes again if they don't improve things. Will try to get production ANALYZE output after these are applied!

Right. Makes sense.

Issue #992, opened 2025-08-28
Make E2E tests usable in CI


binary-koan:
e.g. https://github.com/antiwork/helper/actions/runs/17301995564/job/49115347078?pr=989

* It's way too slow to be useful
* The import-in-the-middle logs are so spammy that GitHub doesn't even load the actual errors
* A bunch of tests are flaky or failing

I think unless we can shorten the runtime to <15 minutes waiting for test runs will slow us down too much and it won't be worth the effort of waiting and fixing broken tests. IIRC the slowdown is mainly specific (common) steps taking forever, so we should try to fix those first to see if it's feasible and worth doing more work on the suite.

Udit-takkar:
> Has there been a performance improvement in these tests? @Udit-takkar

In some tests yes but mostly fixed the flaky/failed E2E tests.

Udit-takkar:
@slavingia Now i am looking into removing all the un necessary tests and combine two tests together if possible

slavingia:
I think 6 min would be considered good for a test run for this kind of app

binary-koan:
Wonder if it would help to build the app up front instead of using the dev server (although it takes ages to build the app as well ...)

binary-koan:
With sharding each e2e action is around 14-18min - still could use improvement, but way less painful than 40min! I still think it should be possible to optimise the app performance in tests as well (when I run them locally I see a lot of 5+ second steps)

sm17p:
@binary-koan can we pause re-factoring work for a while? I am working on running e2e tests with build step. It's gonna end up conflicting

binary-koan:
@Udit-takkar FYI ^

slavingia:
Definitely should be possible to get down to 6 min or so imo for something like helper.

sm17p:
> e.g. https://github.com/antiwork/helper/actions/runs/17301995564/job/49115347078?pr=989
> 
> * It's way too slow to be useful
> * The import-in-the-middle logs are so spammy that GitHub doesn't even load the actual errors
> * A bunch of tests are flaky or failing
> 
> I think unless we can shorten the runtime to <15 minutes waiting for test runs will slow us down too much and it won't be worth the effort of waiting and fixing broken tests. IIRC the slowdown is mainly specific (common) steps taking forever, so we should try to fix those first to see if it's feasible and worth doing more work on the suite.

@slavingia @binary-koan The target of this issue was reached after #1044, right? Can we close this and have another issue for the new target.

PR #1030, opened 2025-09-05
chore: use sharding to run E2E tests on CI

Udit-takkar:
Ref:- https://github.com/antiwork/helper/issues/992

### AI Disclosure:-
I have used Claude Code to plan this task

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -12,8 +12,13 @@ concurrency:
 
 jobs:
   e2e:
-    runs-on: ubicloud-standard-2
     timeout-minutes: 60
+    name: E2E (${{ matrix.shard }}/${{ strategy.job-total }})
+    runs-on: ubicloud-standard-2
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4]
     env:
       CI: true
       OPENAI_API_KEY: test-openai-api-key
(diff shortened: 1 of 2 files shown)


Udit-takkar:
@binary-koan this should help with the time it takes to run all the E2E tests. how can we test this?

binary-koan:
Will hopefully enable running e2e tests automatically on contributor PRs tomorrow so we can see what happens with this!

Udit-takkar:
@Anshgrover23 Okay thanks man.

.github/workflows/e2e.yml
```diff
@@ -55,18 +60,13 @@ jobs:
       - name: Setup E2E test environment
         run: ./scripts/setup-e2e-tests.sh
 
-      - name: Run fail-fast on changed Playwright tests
-        if: github.event_name == 'pull_request'
-        run: ./scripts/e2e.sh playwright test --only-changed=origin/${{ github.base_ref }}
```
binary-koan, line 60: Correct me if I'm wrong, but I don't think we want this - the docs say it runs test files that have changed, but we also want to catch cases where the UI has changed and test files haven't been updated

Udit-takkar:
@binary-koan Yeah it took me a lot of hours to figure out the bug.  We don't have to make any code changes to fix it

